PR #44, opened 2015-08-12
Make avatar a link to profile

stawberri:
Making this a link with an actual url instead of just '#' means that you
can middle click on your avatar in chrome to head over to your profile
in a new tab, without having to use its dropdown!

Assuming I edited your code properly, it shouldn't interfere with
anything. :3

### DIFF
--- a/app/views/layouts/_header.html.haml
+++ b/app/views/layouts/_header.html.haml
@@ -27,7 +27,7 @@
           - unless @user.nil?
             - unless @user == current_user
               %li.hidden-xs{"data-toggle" => "tooltip", "data-placement" => "bottom", title: t('views.actions.group')}
-                %a{href: '#', data: { target: "#modal-group-memberships", toggle: :modal }}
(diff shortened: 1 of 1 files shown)


stawberri:
Really? Oops. I guess I found the wrong link. x.x

pixeldesu:
and the profile picture in the navbar itself shouldn't link to the profile either, as it is the toggle for the dropdown, and if it links to the profile you won't be able to access the dropdown. That's why there is a "View profile" link.

stawberri:
Really? I thought that since it currently links to '#' and your browser doesn't actually navigate there, it wouldn't make a difference what it was a link to, right? :o

It wasn't really meant to be something that you would use often, but if you middle click on your avatar in Chrome right now, all it does is open a new tab to the same page you were on, with an extra # in its url. I thought it'd be more useful if the new tab it opened led to your profile instead! To test it out, I edited the url in chrome's developer thingie, and it really does work as intended--left clicking still opens the dropdown, and middle clicking opens my profile in a new tab. :o

PR #58, opened 2016-06-10
destructive fixing

coderobe:
This change is [Reviewable](https://reviewable.io/reviews/retrospring/retrospring/58)


coderobe:
Manually fixed

PR #63, opened 2016-06-11
Fix screenname changing

twilight-sparkle-irl:
This change is [Reviewable](https://reviewable.io/reviews/retrospring/retrospring/63)


ghost:
it is also too late.

PR #68, opened 2020-04-19
Convert Frontend from Bootstrap 3 to Bootstrap 4

### LINKED ISSUE #3: Frontend Refactoring
Retrosprings frontend was a 2014 hobbyist code project, and that's how the code really looks like at the moment as well.

**To Do:**

**(S)CSS:**  
* [x] Switch from Bootstrap 3 to Bootstrap 4
* [x] Rewrite core elements in a BEM-style syntax
* [x] Use mobile-first approach instead of desktop-first

**JS/Coffee:**
* [x] Try to remove as much jQuery plugins as possible
* [x] Consider moving to a Webpacker pack and using ES6 instead of CoffeeScript

**HAML/Layouts:**  
* [x] Utilize more layouts instead of copying the same partials into base-files
  * A great example for this are all settings and profile pages
* [x] Create more granular templates and a more defined structure than throwing everything into `app/views/shared/`
  * `app/views/answerbox` for all Answerbox partials
  * `app/views/modals` for all Modal elements
  * `app/views/navigation` for all navigations

**Ruby:**
* [x] Refactor a lot of conditionals for classes or URLs into helpers, to reduce in-template logic

pixeldesu:
In order to modernize the frontend, updating Bootstrap to the latest iteration should be done.


pixeldesu:
Styling is mostly done now, I might end up finding a few things while checking through themes and adjusting more things, but here's what's left:

* [x] Refactor current database fields
  * [x] Drop CSS file related fields
  * [x] Remove `default_*` fields
  * [x] Rename `background_text` to `body_text`
  * [x] (SCSS and DB) Rename panel background fields to something more akin "raised" content, because it should be used as a general color for everything that's not in the background.
  * [x] Add a field for muted text
  * [x] Drop `link_color` because in the new theming system, the primary color is the link color.
* [x] Drop old `ThemeHelper#render_theme_with_context`
* [x] Drop theme preview endpoint
* [x] Write specs for new theme helper methods
* [x] Refactor theme code before it actually goes live so it's up-to-standard best practise Ruby code
* [x] Fully implement CSS variable based theming system

pixeldesu:
Gave the Theme settings page a much needed overhaul:

theme page screenshot

pixeldesu:
Ready for review now!

Biggest issue I need to fix is reloading on the theme page causing all color fields to read "#000nan" which is a Turbolinks related issue, which already has been present before.

app/views/settings/_export.haml
```diff
@@ -0,0 +1,32 @@
+.card
+  .card-body
+    %h2 Export your data
+    %p
+      With the announcement of
+      = succeed ',' do
+        %a{href: 'http://blog.retrospring.net/saying-goodbye/'} shutting down Retrospring on June 8 2016
+      we promised you one last feature: exporting all your data.  The data is inside a
```
nilsding: I guess we can get rid of this link + mention of the shutdown.

app/views/tabs/_feed.haml
```diff
@@ -0,0 +1,25 @@
...
+          .p-3
+            %p Looks like you don't have any groups yet!
+            %p
+              You can create groups and add users to them using the
+              %i.fa.fa-fw.fa-users
+              icon in the navigation on user profiles (that are not yours)
+            %p.mb-0
+              Once you have done that, the groups will be listed here and
+              if selected, you'll get a timeline view of all users from
+              that group!
```
nilsding: Use a normal period

app/views/tabs/_feed.haml
```diff
@@ -0,0 +1,25 @@
...
+        - if current_user.groups.empty?
+          .p-3
+            %p Looks like you don't have any groups yet!
+            %p
+              You can create groups and add users to them using the
+              %i.fa.fa-fw.fa-users
+              icon in the navigation on user profiles (that are not yours)
+            %p.mb-0
+              Once you have done that, the groups will be listed here and
+              if selected, you'll get a timeline view of all users from
```
nilsding: when selected

app/views/tabs/_feed.haml
```diff
@@ -0,0 +1,25 @@
...
+        - else
+          Groups
+      .dropdown-menu
+        - if current_user.groups.empty?
+          .p-3
+            %p Looks like you don't have any groups yet!
+            %p
+              You can create groups and add users to them using the
+              %i.fa.fa-fw.fa-users
+              icon in the navigation on user profiles (that are not yours)
```
nilsding: Remove parenthesis, add period at the end

app/views/tabs/_feed.haml
```diff
@@ -0,0 +1,25 @@
...
+    .list-group-item.list-group-item-action.dropdown{class: "#{'active' if @group}"}
+      %a.dropdown-toggle{ type: 'button', data: { toggle: 'dropdown' }, aria: { haspopup: true, expanded: false }}
+        - if @group
+          = @group.display_name
+        - else
+          Groups
+      .dropdown-menu
+        - if current_user.groups.empty?
+          .p-3
+            %p Looks like you don't have any groups yet!
```
nilsding: Use a normal period

config/initializers/assets.rb
```diff
@@ -1,5 +1,13 @@
 # Be sure to restart your server when you modify this file.
 
+Rails.application.config.assets.configure do |env|
```
nilsding: Is this only needed for development/test?  If so, wrap this in a `unless Rails.env.production?`

Alternatively: add `next unless ENV.key?("SPROCKETS_CACHE")` after this line to set this only when `SPROCKETS_CACHE` is present in the environment.

db/migrate/20200504214933_update_theme_fields.rb
```diff
@@ -0,0 +1,75 @@
...
+    change_column_default :themes, :panel_color, 0xF9F9F9
+    change_column_default :themes, :default_color, 0x222222
+    change_column_default :themes, :background_text, 0x222222
+    change_column_default :themes, :background_muted, 0xBBBBBB
+    change_column_default :themes, :background_color, 0xFFFFFF
+    change_column_default :themes, :danger_color, 0xFF0039
+    change_column_default :themes, :warning_color, 0xFF7518
+    change_column_default :themes, :info_color, 0x9954BB
+    change_column_default :themes, :success_color, 0x3FB618
+    change_column_default :themes, :input_color, 0xFFFFFF
```
nilsding: Would say: once the css_* columns are dropped we cannot rollback anymore.  Therefore:

```suggestion
    raise ActiveRecord::IrreversibleMigration
```

spec/helpers/theme_helper_spec.rb
```diff
@@ -0,0 +1,108 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe ThemeHelper, :type => :helper do
+  describe "#get_hex_color_from_theme_value" do
+    it "returns the proper hex value from the theme value" do
+      expect(helper.get_hex_color_from_theme_value(16777215)).to eq("ffffff")
```
nilsding, line 36: Please (also) test this with a value other than `ffffff` -- e.g. just `rgb(0, 0, 255)`

nilsding (changes requested):
Massive.

app/assets/stylesheets/_variables.scss
```diff
@@ -0,0 +1,95 @@
+// Font settings
+@import url(https://fonts.googleapis.com/css?family=Lato:400,700);
```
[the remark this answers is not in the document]
pixeldesu: True that, I actually thought about switching to a System Font Stack, but dropped that idea quickly because some system fonts do be a bad.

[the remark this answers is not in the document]
pixeldesu: Actually, I'd reserve this for a JS refactor, so once we include Webpacker we can pull in Node modules with fonts instead. 

I don't want to put third-party assets into our project nor use `rails-assets` anymore.

app/assets/javascripts/settings.coffee
```diff
@@ -107,19 +107,86 @@ document.head.appendChild previewStyle
 
 previewTimeout = null
 
+$(document).ready ->
```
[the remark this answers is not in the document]
pixeldesu: Actually `$(document).ready` seems to be recommended by `jquery-turbolinks` as it will handle both `ready` and `page:load`?

The issue with this is rooted somewhere different, I believe, but I can't get on why it doesn't work.

I even tried disabling reinitialization with specific checks, but Turbolinks still manages to slip through.

This has been an issue already present on live RS, so I'd also mark that one as a go-to issue for the JavaScript refactor, where we also can consider moving from Turbolinks 2.5 to Turbolinks 5 (!!!), which maybe fixes a lot of our quarrels with it.

app/views/static/about.haml
```diff
@@ -37,24 +37,29 @@
           %h4.heading-about.text-center= t 'views.about.repository.title'
```
nilsding, line 37: (on line 35): looks like the icons are smøl again

nilsding (changes requested):

app/controllers/user_controller.rb
```diff
@@ -110,42 +110,19 @@ def delete_theme
     redirect_to edit_user_profile_path
```
nilsding: I expected a redirect to the edit_user_theme_path

app/views/settings/_theme.haml
```diff
@@ -0,0 +1,135 @@
...
+        .col-sm-12
+          .alert.alert-success
+            A simple success alert—check it out!
+      .row
+        .col-sm-6
+          = f.text_field :dark_color, class: 'color', data: {default: 0x343A40}
+        .col-sm-6
+          = f.text_field :dark_text, class: 'color', data: {default: 0xFFFFFF}
+        .col-sm-12
+          %button.btn.btn-dark.mb-3 A dark button
```
nilsding: These buttons submit the form... is this intentional?

nilsding (changes requested):

nilsding:
When clicking on buttons / tabbing to them (`:active`, I think?), the original active colour is visible:

Peek 2020-05-06 22-26

nilsding:
I think Announcement(s) look out of place (ignore the colours)

Screenshot_20200506_223051

app/views/settings/_theme.haml
pixeldesu: Actually no, pretty weird really. I could turn them into links instead then.

nilsding:
Moderation panel: comments have a bullet point

Screenshot_20200506_223233

app/controllers/user_controller.rb
pixeldesu: Fixed!

nilsding:
More things I noticed:

Not sure if this because of the announcement, but jumbotrons now feature a border.  This is easily noticeable on mobile phones.  
Also: new questions look kinda weird if you're used to the way they are right now, but let's see how long it takes until we're familiar with them.  
One more quick mobile thing: not the entire header is visible -- which is fine by me, but I would've expected the image to be centered.  Seems a bit odd to me.

| IMG_9314 | IMG_9318 |
|-|-|
| IMG_9316 | IMG_9319 |

pixeldesu:
@nilsding 

> Moderation panel: comments have a bullet point

Fixed

> Not sure if this because of the announcement, but jumbotrons now feature a border. This is easily noticeable on mobile phones.

I'll remove that border

> Also: new questions look kinda weird if you're used to the way they are right now, but let's see how long it takes until we're familiar with them.

Actually not intentional and merely an overlook, the unread question design should be the same as before. I need to fix that.

> One more quick mobile thing: not the entire header is visible -- which is fine by me, but I would've expected the image to be centered. Seems a bit odd to me.

I guess due to the header now being a bit higher it's not possible to display it fully there. I'll look into centering it tho.

----

I'll also look into fixing the design of announcements, while I'm at it.

app/views/settings/_theme.haml
```diff
@@ -0,0 +1,135 @@
...
+          .alert.alert-danger
+            A simple danger alert—check it out!
+      .row
+        .col-sm-6
+          = f.text_field :warning_color, class: 'color', data: {default: 0xFFC107}
+        .col-sm-6
+          = f.text_field :warning_text, class: 'color', data: {default: 0xFFFFFF}
+        .col-sm-12
+          .alert.alert-warning
+            A simple warning alert—check it out!
```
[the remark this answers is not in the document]
pixeldesu: True, but that would require another migration for everyone running this already, wouldn't it, since the update-fields one is now a irreversible migration.

app/views/settings/_theme.haml
```diff
@@ -0,0 +1,135 @@
...
+      .row
+        .col-sm-6
+          = f.text_field :raised_background, class: 'color', data: {default: 0xFFFFFF}
+        .col-sm-6
+          = f.text_field :raised_accent, class: 'color', data: {default: 0xF7F7F7}
+    .card-footer
+      %p Some text on top of a accented area on a raised element!
+  .card
+    .card-body
+      %h2 Colors
```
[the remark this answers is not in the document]
pixeldesu: Could reorder it easily, but then the section-naming wouldn't make much sense if "General" came later

nilsding:
More mobile-y things!

<table>
<tbody>
<tr>
<td>

IMG_9320

</td>
<td>

IMG_9321

</td>
</tbody>
</table>

app/views/settings/_theme.haml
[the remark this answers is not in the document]
nilsding: Or does it?  We can keep the values users have set themselves, and change the default for users who have not changed their theme before or wanted a fresh start.

mltnhm:
image

Not sure if my local copy is just borking or if this is an issue

nilsding:
@mltnhm run `rails i18n:js:export`

mltnhm:
image

Can’t exactly read that very well

nilsding:
@mltnhm run `git pull`

raccube:
image

- The time when an answer was submitted gets squished on mobile. 
- On iOS, the Groups dropdown can’t decide how it wants to be styled

raccube:
Upon visiting a themed profile directly, after navigating away, the theme persists.

raccube:
image
On the Discover page

nilsding:
also on the other side of the discover page: Screenshot_20200507_001151

pixeldesu:
> More mobile-y things!

Fixed.

> [Insert Discover page question boxes and user images being large]

Fixed.

> Upon visiting a themed profile directly, after navigating away, the theme persists.

Fixed (and tests adjusted).

> The time when an answer was submitted gets squished on mobile.

Fixed.

> On iOS, the Groups dropdown can’t decide how it wants to be styled

Fixed.

app/views/settings/_theme.haml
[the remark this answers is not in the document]
pixeldesu: Yeah, but changing the default in the database requires a migration.

...or an adjustment to the main one, which wouldn't apply it to people testing it right now.

pixeldesu:
> I think Announcement(s) look out of place (ignore the colours)

Fixed.

> Also: new questions look kinda weird if you're used to the way they are right now, but let's see how long it takes until we're familiar with them.

Fixed.

> Not sure if this because of the announcement, but jumbotrons now feature a border. This is easily noticeable on mobile phones.

Fixed.

pixeldesu:
@nilsding @mltnhm Please re-review with latest changes. Most stuff you reported is addressed now.

raccube:
image
Could do with some more spacing here.

raccube:
image

raccube:
image
Feel like this could look a bit nicer - the bell icon could be larger.

raccube:
image

pixeldesu:
@mltnhm Fixed all issues!

app/assets/stylesheets/_variables.scss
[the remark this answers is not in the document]
nilsding: Also fine, as long as we'll get rid of it really soon.

nilsding (approved):

PR #72, opened 2020-04-20
Set up GitHub Actions

### LINKED ISSUE #71: Add GitHub Actions for CI
Should only run RSpec for now.

raccube:
Baby's first GitHub Actions pipeline.

Closes #71

### DIFF
--- a/.github/workflows/retrospring.yml
+++ b/.github/workflows/retrospring.yml
@@ -20,6 +20,10 @@ jobs:
     services:
       postgres:
         image: postgres:10.12
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: justask_test
         ports:
           - 5432/tcp
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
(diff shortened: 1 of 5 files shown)


.github/workflows/retrospring.yml
```diff
@@ -0,0 +1,75 @@
...
+        env:
+          RAILS_ENV: test
+          POSTGRES_PORT: ${{ job.services.postgres.ports[5432] }}
+          POSTGRES_HOST: localhost
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: justask_test
+      - name: Run tests
+        run: |
+          bundle exec rake db:setup
```
nilsding: Isn't this redundant as this is run before anyway?

.github/workflows/retrospring.yml
```diff
@@ -0,0 +1,75 @@
...
+          bundle exec rake spec
+        env:
+          RAILS_ENV: test
```
nilsding: hmm, I wonder if these env vars could be specified just once.  Either way, it's fine for now.

nilsding (approved):

.github/workflows/retrospring.yml
raccube: Oops. Forgot to drop that.

PR #77, opened 2020-04-22
Use email field for password reset

raccube:


pixeldesu (changes requested):
I'm conflicted, because this _technically_ removes the ability to use either the screen_name or email for recovering the account.

Thinking about it, actually..._both_ and `screen_name` only are a bad choice, because these can be used to eventually spam someone. A screen name is public, so you can eventually go and send a ton of emails to someone.

It'd probably be a sensible choice to just switch password recovery to only email instead.

raccube:
Just so we have it here:
image

app/views/devise/passwords/new.html.haml
```diff
@@ -4,7 +4,7 @@
   = bootstrap_form_for(resource, as: resource_name, url: password_path(resource_name), html: { method: :post }) do |f|
     = devise_error_messages!
 
-    = f.text_field :login, autofocus: true, label: "User name"
+    = f.text_field :email, autofocus: true, label: "Email address"
```
nilsding: Hmm, wonder if this renders as an email field (with browser-side validation)

nilsding (approved):

app/views/devise/passwords/new.html.haml
raccube: Now it does.

pixeldesu (approved):

PR #82, opened 2020-04-25
Fix exception being thrown upon attempting to ban a user

### LINKED ISSUE #75: Banning over the frontend doesn't work anymore
Not the functionality in the backend, but the modal on user profiles appears to have lost functionality.

My first guess seems to be that the IDs `bootstrap_forms` assigns to elements has changed and with that, the CoffeeScript related to displaying data and sending the AJAX request is now broken.

raccube:
Fixes #75

### DIFF
--- a/app/assets/javascripts/moderation/ban.coffee
+++ b/app/assets/javascripts/moderation/ban.coffee
@@ -1,8 +1,8 @@
 load = ->
+  return unless document.getElementById('ban-control-super') != null
   modalEl = $("#modal-ban")
   modalEl.modal "hide"
   modalForm = modalEl.find("form")[0]
-  return unless document.getElementById('ban-control-super') != null
   banCheckbox = modalForm.querySelector('[name="ban"][type="checkbox"]')
   permabanCheckbox = modalForm.querySelector('[name="permaban"][type="checkbox"]')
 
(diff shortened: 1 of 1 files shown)


nilsding (approved):

app/assets/javascripts/moderation/ban.coffee
```diff
@@ -1,22 +1,27 @@
...
 
-  parent.find('#_ban').on "change", (event) ->
+  banCheckbox.addEventListener "change", (event) ->
     $t = $ this
     if $t.is(":checked")
       $("#ban-controls").show()
     else
       $("#ban-controls").hide()
-  parent.find('#_permaban').on "change", (event) ->
+  permabanCheckbox.addEventListener() "change", (event) ->
```
pixeldesu:
```suggestion
  permabanCheckbox.addEventListener "change", (event) ->
```

Not even sure if this qualifies as a syntax error in CoffeeScript's book, but at least choose to keep it same-y. Removing the parenthesis would adjust it to the same style as all the other `addEventListener` calls you made.

pixeldesu (changes requested):

pixeldesu (approved):

PR #87, opened 2020-04-25
Move from paperclip to Active Storage

### LINKED ISSUE #85: Replace paperclip with maintained alternative
[Paperclip](https://github.com/thoughtbot/paperclip) has been deprecated.

So, for file uploads and processing, we should switch to something that is actually maintained.

Two alternatives for this are:
- [Rails' `ActiveStorage`](https://guides.rubyonrails.org/active_storage_overview.html)
- [CarrierWave](https://github.com/carrierwaveuploader/carrierwave)

raccube:
Closes #85


raccube:
Use CarrierWave instead. Active Storage doesn’t work nicely for our use case.

PR #101, opened 2020-05-09
General housekeeping

pixeldesu:
* Moves some of the files GitHub works with into `.github`
* Added `CODEOWNERS` file so people responsible for specific parts of code get automatically requested for review
* Improved Docker setup
  * Removed deprecated `links` keyword and replaced it with `depends_on`
  * Use Node 12 instead of Node 6
  * Condensed the package installation into a single package to speed up installation
  * Added entrypoint that clears existing PIDs on startup
* Removed `TODO` file from repository as most things either are obsolete or already have issues created for them


.docker/ruby/Dockerfile
```diff
@@ -1,30 +1,26 @@
...
+RUN apt-get update -qq && \
+    apt-get install -y --no-install-recommends build-essential \
+    libpq-dev postgresql-client \ 
+    libxml2-dev libxslt1-dev \
+    libmagickwand-dev imagemagick \
+    nodejs
 
-# for nokogiri
-RUN apt-get install -y libxml2-dev libxslt1-dev
+RUN rm -rf /var/lib/apt/lists/*
```
nilsding: guess these steps could be combined

nilsding (approved):

.docker/ruby/Dockerfile
pixeldesu: Done!

PR #108, opened 2020-05-11
Fix notification menu clipping out of view

### LINKED ISSUE #96: Notification texts are cut off on notification page
Reported here: https://retrospring.net/xvilyv/a/169642

I'm not even sure if that was a thing before or not, but probably a good idea to show full texts on the notification list if possible. It should only be shortened in the notification dropdown.

pixeldesu:
Pretty self-explanatory. Fixes #96 since that is the issue it was actually reported for.


raccube (approved):
:shipit: LGTM :shipit:

Issue #74, opened 2020-04-21
Translation Refactoring


pixeldesu:
As it pretty much was one of the first larger projects we have done with localization, it's pretty broken.

We're using split translations based on English sentence order and base vocabulary drop-ins without proper pluralization, causing most languages aside of English to look pretty bad.

To fix this, pretty much all of the translations have to be redone.

**To Do:**
* [x] Remove all locales aside of English
* [ ] Refactor english translations to use placeholders and pluralization properly
* [ ] Setup Crowdin for new translations in evaluated languages

PR #120, opened 2020-05-25
Rename Groups to Lists

### LINKED ISSUE #52: i am sometimes receiving questions from people i do not follow
i'm not following Jack_Kerouac because i have zero interest in interacting with them, but [this question](https://retrospring.net/Jack_Kerouac/q/249784) still showed up in my inbox

perhaps this is intentional and i should hold out for #12, but if it is i'm not sure what 'following' even does

i also have a friend who has reported that this is why they stopped using the site months ago

### LINKED ISSUE #113: Groups to Lists
The current implementation of Groups in Retrospring aren't really groups, they are more akin to something like Twitters List feature, where you group people for yourself.

An issue about this is the fact that this also allows you to ask questions to those groups, without the consent of the user specifically, which is really bad user experience (and with that, probably the cause for issues like #52)

This worked out for so long because we made misconceptions about this feature, and barely anyone ever used it, until we now probably made it a bit more obvious that it exists.

**To Do:**
* Rename every `Group*` reference to `List*`
* Remove ability to send questions to group/list members

nilsding:
Resolves #113, closes #52


app/views/static/front.haml
```diff
@@ -65,11 +65,10 @@
...
     .col-md-4.col-sm-4.col-xs-12
       %h3
-        %i.fa.fa-fw.fa-users.text-primary
-        Groups
+        %i.fa.fa-fw.fa-list.text-primary
+        Lists
       %p
-        Want to ask a question to a specific set of people? No problem!
-        Set up unique groups of users and send them questions apart from everyone else.
+        ??????????
```
raccube: This should be filled or removed before merging.

raccube (changes requested):
Some of the locales still reference groups. (cs, it, ja) Guess we would need translations for that.

app/views/static/front.haml
nilsding: Maybe @pixeldesu has an idea what to place there now.

pixeldesu:
> Some of the locales still reference groups. (cs, it, ja) Guess we would need translations for that.

@mltnhm none of those locales are used by active users.

raccube:
@pixeldesu Makes sense. As this isn't a popular feature and neither are those locales; we can leave it until the translation refactor (#74)

app/views/static/front.haml
pixeldesu: _"Want to keep an oversight of things? Organize users in Lists to create timelines of their answers."_

Or something along those lines.

nilsding: Sounds good.  I'll take that

raccube (approved):

pixeldesu (approved):
I'm amazed that even with mostly regex-replacing, this still worked so well. Tested it locally. LGTM 👍

nilsding:
Well, there was this one wrong selector though.  ;-)

PR #134, opened 2020-05-31
Drop DejaVu Sans from sans-serif fonts

### LINKED ISSUE #132: (Color) emoji rendering inconsistent
`font-family` includes `DejaVu Sans`, which contains an incomplete set of emoji, some of which are not available in color. Systems which have `DejaVu Sans` installed will grab the emoji from there, causing inconsistent rendering:
image
image
None of the other fonts in `font-family` provide emoji coverage, thus removing `DejaVu Sans` from the list of fonts falls back to the system sans-serif fallback, rendering the emoji properly:
image
image

This issue affects any system which has DejaVu Sans available, and causes _some_ (but not all) emoji to be bad.
I would just drop DejaVu Sans from the fonts. as a quick fix - it's not high up in the fallbacks anyways, and is not currently providing any other characters as those are covered by other fonts in the list.

coderobe:
Fixes #132

### DIFF
--- a/app/assets/stylesheets/_variables.scss
+++ b/app/assets/stylesheets/_variables.scss
@@ -1,7 +1,7 @@
 // Font settings
 @import url(https://fonts.googleapis.com/css?family=Lato:400,700);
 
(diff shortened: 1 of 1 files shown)


pixeldesu (approved):
LGTM 👍

PR #135, opened 2020-05-31
Add dropdown toggle to list item as well

### LINKED ISSUE #131: "Lists" tab on Timeline page not clickable everywhere
image
The tabs "Timeline" and "Public" can be clicked anywhere on the tab field, "Lists" can only be clicked on the text and the arrows.

coderobe:
Fixes #131

### DIFF
--- a/app/views/tabs/_feed.haml
+++ b/app/views/tabs/_feed.haml
@@ -3,7 +3,7 @@
     = list_group_item t('views.general.timeline'), root_path
     - if APP_CONFIG.dig(:features, :public, :enabled)
       = list_group_item t('views.general.public'), public_timeline_path
-    .list-group-item.list-group-item-action.dropdown{ class: list ? 'active' : '' }
+    .list-group-item.list-group-item-action.dropdown{ class: list ? 'active' : '', data: { toggle: :dropdown } }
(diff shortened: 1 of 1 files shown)


coderobe:
The side-effect of this is that the dropdown now looks like this:
image

Prior to this change, it looked like this: (it's offset to the tab title)
image

pixeldesu (changes requested):
Conflicted, because markup-wise there are now two `data-toggle="dropdown"` elements nested, and Bootstrap generally expects the toggle and menu being aside of each other. _Or so I thought until now._

This way the menu now aligns with the list group item, and with the text content it fills out the full width of it. If you have shorter list names, it doesn't:

Example:


Generally I'm okay with the toggle being on the `list-group-item`, but then you should remove it from the dropdown toggle (and also move the other `aria-*` attributes to the list group item.

Then, you should probably add an override styling for `dropdown-menu` classes in `app/assets/stylesheets/overrides/_dropdown.scss`, so that the menus in `list-group-item-action` elements are always in full-width.

```scss
.list-group-item-action .dropdown-menu {
  width: 100%;
}
```

pixeldesu:
Fixed this in #262

PR #136, opened 2020-05-31
Use display name instead of screen name in smile tooltip

### LINKED ISSUE #127: User name / Display name UI inconsistencies
Questions get shown with the display name of the asker
image

When hovering over the avatar in the "smiles" section, however, the hover text displays their username
image

coderobe:
Fixes #127

### DIFF
--- a/app/views/answerbox/_smiles.haml
+++ b/app/views/answerbox/_smiles.haml
@@ -7,6 +7,6 @@
     - else
       - a.smiles.all.each do |smile|
         %a{ href: show_user_profile_path(smile.user.screen_name),
-          title: smile.user.display_name,
+          title: user_screen_name(smile.user, anonymous: false, url: false),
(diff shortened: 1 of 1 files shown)


app/views/answerbox/_smiles.haml
```diff
@@ -7,6 +7,6 @@
     - else
       - a.smiles.all.each do |smile|
         %a{ href: show_user_profile_path(smile.user.screen_name),
-          title: smile.user.screen_name,
+          title: smile.user.display_name,
```
pixeldesu: What if the display name is empty tho?

Use the designated helper `UserHelper#user_screen_name` for this: 
https://github.com/Retrospring/retrospring/blob/master/app/helpers/user_helper.rb

```suggestion
          title: user_screen_name(smile.user, url: false),
```

pixeldesu (changes requested):

pixeldesu (approved):

PR #137, opened 2020-05-31
Add bottom padding to userbox col in show_follow

### LINKED ISSUE #116: User lists don't have spacing between rows of users
There is no vertical spacing between users in the follow/following as well as moderation user list.

**Example:** https://retrospring.net/pixeldesu/followers

coderobe:
Fixes #116

image

Before this patch:
image

### DIFF
--- a/app/views/moderation/priority.haml
+++ b/app/views/moderation/priority.haml
@@ -6,5 +6,5 @@
 #users
   .row.row-cols-1.row-cols-sm-2.row-cols-md-3
     - @users.each do |u, c|
-      .col
+      .col.pb-3
         = render 'moderation/userbox', user: u, count: (c || 0)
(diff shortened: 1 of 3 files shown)


coderobe:
🧙 Whoosh
✨Nobody saw the ninja-commit✨

pixeldesu (changes requested):
~This just adds the padding to users rendered by the pagination, they still miss from the initially rendered one:
https://github.com/Retrospring/retrospring/blob/master/app/views/user/show_follow.haml#L3~

If you're at it, you might add the padding to the user list in the moderation section as well (it's practically identical):
https://github.com/Retrospring/retrospring/blob/master/app/views/moderation/priority.haml#L9

pixeldesu (approved):

PR #146, opened 2020-10-03
Docker Tweaks

raccube:
* Updated the docker-compose config to store Postgres data to prevent it from being lost
* Updated container build to use all available CPU threads for `bundle install`


pixeldesu (approved):
LGTM 👍

PR #147, opened 2020-10-18
Implement Two Factor Authentication

### LINKED ISSUE #84: Implement 2FA
I don't think I have to explain this one.

raccube:
Closes #84


raccube:
image
Setup UI

app/controllers/user/sessions_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+        session[:user_sign_in_uid] = resource.id
+        sign_out(resource)
+        redirect_to user_two_factor_entry_url
+      else
+        if resource.authenticate_otp(params[:user][:otp_attempt])
+          continue_sign_in(resource, resource_name)
+        else
+          sign_out(resource)
+          flash[:error] = t('devise.failure.invalid')
+          redirect_to root_url
```
nilsding: not sure if redirecting to the root_url is a nice thing to do; maybe a redirect to the sign in path would be better here?

app/controllers/user/sessions_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      session.delete(:user_sign_in_uid)
+    else
+      self.resource = warden.authenticate!(auth_options)
+    end
+
+    if resource.active_for_authentication? && resource.otp_module_enabled?
+      if params[:user][:otp_attempt].blank?
+        session[:user_sign_in_uid] = resource.id
+        sign_out(resource)
+        redirect_to user_two_factor_entry_url
```
nilsding: No need for this workaround with the second page, just lock Warden for this request after the `sign_out` so `current_user` and the likes do not sign the user in again. (this is due to `current_user` calling `warden.authenticate`, which performs a sign in for the user again due to the parameters for the database authenticatable stuff being present)

```suggestion
        warden.lock!
        render 'auth/two_factor_authentication'
```

--> then get rid of `def two_factor_entry` and the route

app/controllers/user/sessions_controller.rb
```diff
@@ -0,0 +1,47 @@
+class User::SessionsController < Devise::SessionsController
+  def create
```
nilsding: override the `#new` method where `:user_sign_in_uid` is deleted from the session (and `super` is called) to avoid forcing entry of the OTP key for a different user during sign ins; also adapt the `routes.rb` for that

app/views/settings/security/_totp_setup.haml
```diff
@@ -0,0 +1,42 @@
+.totp-setup
```
nilsding: this page looks odd on phones at the moment

db/migrate/20201001172537_add_otp_secret_key_to_users.rb
```diff
@@ -0,0 +1,10 @@
+class AddOtpSecretKeyToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :otp_secret_key, :string
+    add_column :users, :otp_module, :integer, default: 0, null: false
+
+    User.find_each do |user|
+      user.update_attribute(:otp_secret_key, User.otp_random_secret)
+    end
```
nilsding: please check if this is really required

spec/controllers/user/sessions_controller_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+
+describe User::SessionsController do
+  before do
+    @request.env["devise.mapping"] = Devise.mappings[:user]
+  end
```
nilsding: ???

please add a comment explaining why this is required if this is required

nilsding (changes requested):

app/controllers/user/sessions_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def two_factor_entry
+    unless session.has_key? :user_sign_in_uid
+      redirect_to root_url
+      return
+    end
+
+    self.resource = User.find(session[:user_sign_in_uid])
+    render 'auth/two_factor_authentication'
+  end
```
nilsding: not required anymore

config/routes.rb
```diff
@@ -48,7 +48,8 @@
   as :user do
     # :sessions
     get 'sign_in' => 'devise/sessions#new', as: :new_user_session
```
nilsding: this now needs to be `user/sessions#new` -- otherwise the memorised uid in the session is never removed

nilsding (changes requested):

nilsding (approved):

raccube:
## TOTP Setup screenshots
Desktop:
image
Mobile:
image
Tablet:
image
Tablet (portrait):
image

pixeldesu (approved):
Generally, superb execution on the frontend side, nothing I feel I have to complain about, looking over it. Layout seems fine too.

The fact that the QR code adapts to the theme is awesome, but I fear that might cause it to become unreadable with certain themes, especially lighter ones, doesn't it? Generally not much of a problem, and we still can add a disclaimer later on.

config/justask.yml.example
```diff
@@ -68,3 +68,6 @@ hcaptcha:
   enabled: false
   site_key: ''
   secret_key: ''
+
+# TOTP Drift period in seconds
+otp_drift_period: 30
```
nilsding, line 73: change deployed

nilsding (approved):

raccube:
Won't do another deployment until recovery codes are implemented.

PR #109, opened 2020-05-11
Refactor CoffeeScript into TypeScript

pixeldesu:
This PR is part of the ongoing effort to refactor the frontend of Retrospring, it handles the JavaScript part of the frontend, which is moving from Rails-handled CoffeeScript to Webpacker-built ES6 with packs.

It'll also introduce the effort to remove as many of the `rails-assets` and jQuery gems currently in use, with more modern alternatives.

**Roadmap of JS features:**
* [ ] Pagination functionality
* Inbox
  * [x] Generating questions
  * [x] Deleting all questions
  * [ ] Deleting all author questions
  * [x] Searching for an author
  * [x] Answering a single question
  * [x] Deleting a single question
  * [ ] Opening options on questions
  * [ ] Keyboard shortcuts
* Answerboxes
  * [ ] Opening the comment section
  * [ ] Adding a new comment
  * [ ] Removing comments
  * [ ] (Un)smiling comments
  * [ ] Reporting comments
  * [ ] Comment character count
  * [ ] Removing an answer
  * [ ] Reporting an answer
  * [ ] (Un)smiling an answer
  * [ ] (Un)subscribing to answer updates
* Questionboxes
  * [ ] Asking all followers/a group questions
  * [ ] Removing a question
  * [ ] Reporting a question
  * [ ] User page specific questionbox functions (user promotion etc.)
  * [ ] Keyboard shortcuts
* Moderation
  * [ ] Banning a user
  * [ ] Removing a report
  * [ ] Voting on a report
  * [ ] Commenting on a report
  * [ ] Setting user privileges
* Groups
  * [ ] Adding new groups
  * [ ] Adding/removing members of groups
  * [ ] Removing groups
  * [ ] Keyboard shortcuts
* Questions
  * [ ] Answering question (non-inbox, share logic with Inbox here tho)
  * [ ] Keyboard shortcuts
* Settings
  * [ ] Profile picture cropping
  * [ ] Header cropping
  * [ ] Theme preview
* User
  * [ ] (Un)following user
  * [ ] Reporting user

PR #155, opened 2020-12-24
Setup Webpacker

pixeldesu:
This is what #109 should have been, but the waaaaay too large scope kinda killed it. We'll never finish it that way.

This PR just sets up Webpacker and CoffeeScript, trying to port as much over without touching the actual code, so we can later refactor it to TS and remove jQuery and such.


pixeldesu:
Most of the stuff works with Webpacker now.

I'm still encountering a few issues here and there, but most of it, including jQuery, works fine. It's almost 4 AM, so I'll take a break now, but this will probably be done rather quickly.

pixeldesu:
Test runs fixed. For "some" reason `coffee-loader` does not work with Webpack 5, so I had to drop the Webpack version to 4 and `coffee-loader` to 1.0.1

I love frontend.

raccube (approved):

PR #156, opened 2020-12-25
Add sign in tests

raccube:


spec/controllers/user/sessions_controller_spec.rb
```diff
@@ -21,5 +21,53 @@
 
       expect(subject).to have_rendered('auth/two_factor_authentication')
     end
+
+    context "2fa sign in attempt" do
+      subject {
```
nilsding: nit: multi-line blocks should be done with `do ... end`

spec/controllers/user/sessions_controller_spec.rb
```diff
@@ -21,5 +21,53 @@
...
+        it "redirects to the timeline" do
+          expect(subject).to redirect_to :root
+        end
+      end
+
+      context "correct recovery code" do
+        let(:code_input) { 'raccoons' }
+
+        it "consumes the recovery code" do
+          user.totp_recovery_codes << TotpRecoveryCode.create(code: 'raccoons')
```
nilsding: would place this in a `before` block as this is part of the context setup

nilsding (approved):
Looking good overall

PR #167, opened 2020-12-26
Add Typescript support

### LINKED ISSUE #158: Add TypeScript setup to Webpacker
For the general refactoring of the JS portion of the frontend, we want to switch to TypeScript instead of CoffeeScript.

For the meantime, we need to support both, so the TS compilation should not replace the CoffeeScript one.

**To Do:**
* [x] TypeScript compilation support has been added to the project
* [x] General TypeScript stuff has been bootstrapped

**Note:** For some general starting stuff, code from the old PR #109 can still be looked at in the [feature/webpacker](https://github.com/Retrospring/retrospring/tree/feature/webpacker) branch.

pixeldesu:
Adds #158 and opens up the ability to slowly migrate the features from the other issues over to Typescript.


app/javascript/retrospring/initializers/bootstrap.ts
```diff
@@ -0,0 +1,15 @@
+import 'bootstrap';
+import $ from 'jquery';
+
+/**
+ * This module sets up Bootstraps JavaScript
```
raccube: nit
```suggestion
 * This module sets up Bootstrap's JavaScript
```

raccube (approved):

pixeldesu:
I'll probably merge this as-is, but we need to remember something (especially @nilsding).

Yarn needs to be present on the RS host otherwise Webpacker will have a fun party, especially with the previous PR already, we need to ready the deployment for that.

raccube:
What do we need to run when deploying?

pixeldesu:
> What do we need to run when deploying?

```shell
$ yarn install --frozen-lockfile
```

To ensure the packages are present before `assets:precompile` is run.

PR #173, opened 2021-04-25
Fix broken services integration after omniauth bump

idlechara:
The last merged commit (45cb0cab26a818094767352d554e546ae22314b3) bumped omniauth to version `1.2.0` during the gemfile update breaking service integration for both Tumblr and Twitter.

https://github.com/Retrospring/retrospring/blob/45cb0cab26a818094767352d554e546ae22314b3/Gemfile.lock#L329

The cause is that omniauth `1.2.0` requires the gem `omniauth-rails_csrf_protection` and also forces the request to use only post methods as described here: https://stackoverflow.com/questions/66009147/no-route-matches-get-auth-twitter-omnia.

This commit adds the missing gem and changes the HTTP request used to link the service via omniauth to use POST instead of GET.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -97,3 +97,5 @@ group :development, :test do
 end
 
 gem "webpacker", "~> 5.2"
+
+gem "omniauth-rails_csrf_protection", "~> 1.0"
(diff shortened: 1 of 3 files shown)


nilsding (approved):
Looks good to me, thank you for the contribution!

PR #175, opened 2021-05-16
Upgrade to Bootstrap 5

pixeldesu:
We can _finally_ drop jQuery with this.

**To Do:**
* [ ] Migrate as much as required by the [migration guide](https://getbootstrap.com/docs/5.0/migration/)
  * [ ] Quite specifically, `.media` was removed, which is used a lot in answer/questionboxes
* [ ] Fix styling, where it is broken
* [ ] Fix broken JS, and maybe, if the effort is low, port it to TS
* [ ] Drop jQuery entirely, which also means replacing some dependencies and functional modules we use


pixeldesu:
_This branch is 3 commits ahead, 428 commits behind master._

PR #182, opened 2021-08-05
Add links for jumping to rails admin on answers, questions and profiles

raccube:


app/views/answerbox/_actions.haml
```diff
@@ -38,3 +38,7 @@
         %a.dropdown-item{ href: '#', data: { a_id: a.id, action: 'ab-report' } }
           %i.fa.fa-exclamation-triangle
           = t 'views.actions.report'
+      - if current_user.has_role? :administrator
+        %a.dropdown-item{ href: rails_admin_path + "/answer/#{a.id}", target: '_blank' }
```
nilsding: just `"#{rails_admin_path}/answer/#{a.id}"` would be fine too

Maybe there's even a more special helper that you can use like `rails_admin_path(answer: a)`

app/views/user/_actions.haml
```diff
@@ -28,3 +28,7 @@
             %a.dropdown-item{ href: '#', data: { target: '#modal-ban', toggle: :modal } }
               %i.fa.fa-ban
               = t 'views.actions.ban'
+        - if current_user.has_role? :administrator
+          %a.dropdown-item{ href: "#{rails_admin_path}/user/#{user.id}", target: '_blank' }
```
pixeldesu: While those look fine now, couldn't we make it a bit cleaner?

Like, having a helper generate the full URL, being in parts the `rails_admin_path` but figuring the rest out itself? Working out the other URL fragment is the `resource` in some format + it's id

So having something like `rails_admin_path_for_resource(user)` would be swell, so we don't have a bunch of concatenation in the templates themselves!

pixeldesu (changes requested):

pixeldesu (approved):
LGTM, but maybe fix the code coverage

PR #183, opened 2021-08-05
Add share buttons to answers using Web Share API

raccube:
Example:
Screenshot of change


app/assets/stylesheets/components/_answerbox.scss
```diff
@@ -61,6 +61,20 @@
...
+      justify-content: space-between;
+      padding-top: 10px;
+    }
+  }
+
+  @include media-breakpoint-down('md') {
+    .card-body {
+      padding-bottom: .6rem;
+    }
+  }
```
pixeldesu: mobile first, no downwards queries please. Use `media-breakpoint-up` instead. And inverse styling (having styles on mobile, reset on other queries if necessary)

app/javascript/retrospring/initializers/share.ts
```diff
@@ -0,0 +1,20 @@
...
+                    url: card.find('.answerbox__answer-date a')[0].href
+                }).then(() => {
+                    // do nothing, prevents exception from being thrown
+                }).catch(() => {
+                    // do nothing, prevents exception from being thrown
+                })
+            })
+        }
+    })
+}
```
pixeldesu: * Initializers are meant to set up larger global features (Bootstrap, jQuery, etc. pp.), this is a single feature.
* The code doesn't need to be initialized if no `button[name=ab-share]` is present (save code execution)
* **Don't use jQuery!** It's planned to be removed (and it already is in a branch)

You can use the [`feature/webpacker` branch](https://github.com/Retrospring/retrospring/tree/feature/webpacker/app/javascript) as a reference on how I implemented features before!

pixeldesu (changes requested):

app/javascript/packs/application.ts
```diff
@@ -1,3 +1,7 @@
 import start from '../retrospring/common';
+import initAnswerbox from '../retrospring/features/answerbox/index';
 
 start();
+document.addEventListener('turbolinks:load', initAnswerbox);
+document.addEventListener('turbolinks:load', initAnswerbox);
+document.addEventListener('turbolinks:load', initAnswerbox);
```
pixeldesu: Eh?

app/javascript/retrospring/features/answerbox/index.ts
```diff
@@ -0,0 +1,20 @@
+import registerEvents from "retrospring/utilities/registerEvents";
+import {createShareEvent} from "./share";
+
+export default (): void => {
+  if (!('share' in navigator)) {
+    return;
+  }
```
pixeldesu: Good, but this also breaks setting any `answerbox` related functionality if `navigator.share` is not there, maybe move this check to `createShareEvent`.

app/javascript/retrospring/features/answerbox/share.ts
```diff
@@ -0,0 +1,11 @@
+export function createShareEvent(answerbox: HTMLElement): () => void {
+  return function (): void {
+    navigator.share({
+      url: answerbox.querySelector<HTMLAnchorElement>('.answerbox__answer-date > a').href
+    }).then(() => {
+      // do nothing, prevents exce ption from being thrown
+    }).catch(() => {
+      // do nothing, prevents exception from being thrown
+    })
```
pixeldesu: Hmm, this looks kinda ugly, we could just do an empty function as a base `noop` instead.

Maybe create one in `retrospring/utilities`

```suggestion
    })
    .then(noop)
    .catch(noop)
```

pixeldesu (changes requested):

app/javascript/packs/application.ts
raccube: not sure how I managed that one

app/javascript/retrospring/utilities/noop.ts
```diff
@@ -0,0 +1,3 @@
+export default () => void {
```
pixeldesu:
```suggestion
export default (): void => {
```

pixeldesu: I'm actually suprised that's not a syntax error

pixeldesu (approved):

PR #185, opened 2021-08-06
Add OpenGraph metadata on user profiles and answers

raccube:


pixeldesu (changes requested):
Generally better approach:

Give entities an `opengraph/twitter` method that returns the fields and have the generic helper that creates the tags take only the entity, looking for the methods existing to get the mappings from there instead.

So stuff is where it makes sense (entity->graph mapping in entity itself, and not the helper code)

raccube:
I tried to implement it that way but ran in to a problem as these values include a bunch of URLs which we get from route constants. These constants are only accessible in controllers. So we'd either have to reimplement those or generate a bunch of extra cruft to pass them through.

pixeldesu (approved):
LGTM

PR #187, opened 2021-08-11
Show question link for answered questions with only one answer

### LINKED ISSUE #43: you can't get to questions that are not in your inbox that only have one answer
it seems that there's no way to get to the detail page for questions from people you don't follow that only have one answer, even if they were asked of everyone; where most questions would have a '_n_ answers' link, if there's only one, no such link exists. even when you're looking at the person's list of questions, you can't see the answers if only one exists

raccube:
Closes #43

### DIFF
--- a/app/controllers/ajax/question_controller.rb
+++ b/app/controllers/ajax/question_controller.rb
@@ -32,7 +32,8 @@ class Ajax::QuestionController < AjaxController
     begin
       question = Question.create!(content: params[:question],
                                   author_is_anonymous: is_never_anonymous ? false : params[:anonymousQuestion],
-                                  user: current_user)
+                                  user: current_user,
+                                  direct: params[:rcpt] != 'followers')
     rescue ActiveRecord::RecordInvalid => e
       NewRelic::Agent.notice_error(e)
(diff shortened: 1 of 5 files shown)


raccube:
Maybe we should add a `direct` field to questions that were sent directly to a user rather than to followers

nilsding:
Yes, please add that field.  This is the reason why the check for more than one answer was in there.

On the question page, you should also hide the "Answer this question" box at the bottom of the page if it's a direct question.

app/controllers/ajax/question_controller.rb
```diff
@@ -32,7 +32,8 @@ def create
     begin
       question = Question.create!(content: params[:question],
                                   author_is_anonymous: is_never_anonymous ? false : params[:anonymousQuestion],
-                                  user: current_user)
+                                  user: current_user,
+                                  direct: params[:rcpt] != 'followers')
```
nilsding, line 36: Please add some specs for checking the direct flag

pixeldesu (approved):

nilsding (approved):

PR #193, opened 2021-08-13
Use `break-word` instead of `break-all`

raccube:
This will prevent words being broken mid-word


raccube:
I wrote this over SSH on my phone.

pixeldesu (approved):

PR #194, opened 2021-08-14
Implement ban history

### LINKED ISSUE #140: Use sane date format for specifying ban duration
Right now it uses `%m/%d/%Y %I:%M %p`, which shouldn't even be a thing.  Use ISO 8601 date format instead.

Maybe also try out how `Time.parse` from `require "time"` parses certain strings -- perhaps we don't need a specific format then.

https://github.com/Retrospring/retrospring/blob/5ae5ed6c8dd6f1b150f19543574a63069f46313a/app/controllers/ajax/moderation_controller.rb#L118

raccube:
Closes #140 

- [x] Tests
- [x] Update Rake tasks
- [x] Update Rails Admin
- [x] Frontend: Unban

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -228,7 +228,7 @@ class User < ApplicationRecord
   end
 
   def unban
-    self.user_bans.current.update(expires_at: DateTime.now)
+    self.user_bans.current.update!(expires_at: DateTime.now)
   end
 
   # Bans a user.
@@ -237,9 +237,6 @@ class User < ApplicationRecord
   # @param reason [String] Reason for the ban. This is displayed to the user.
   # @param banned_by [User] User who instated the ban
   def ban(duration, duration_unit = 'hours', reason = nil, banned_by = nil)
-    raise Errors::InvalidBanDuration unless %w[hours days weeks months].include? duration_unit
-
-    expiry = duration && DateTime.now + duration.public_send(duration_unit)
     self.user_bans.create(expires_at: expiry, reason: reason, banned_by: banned_by)
   end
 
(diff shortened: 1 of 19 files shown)


pixeldesu (approved):
LGTM 👍

app/models/user.rb
```diff
@@ -218,21 +223,31 @@ def report_comment(report, content)
...
   def unban
-    self.update(permanently_banned: false, ban_reason: nil, banned_until: nil)
+    UseCase::User::Unban.call(id)
   end
 
-  def ban(buntil=nil, reason=nil)
-    if buntil == nil
-      self.update(permanently_banned: true, ban_reason: reason)
+  # Bans a user.
+  # @param duration [Fixnum, nil] Ban duration
```
nilsding: Fixnum doesn't really exist these days, `Integer` is where it's at
```suggestion
  # @param duration [Integer?] Ban duration
```

spec/models/user_ban_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe UserBan, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
nilsding: Delete this

nilsding (changes requested):

spec/models/user_ban_spec.rb
raccube: 669aa579

app/models/user.rb
raccube: 669aa579

nilsding (approved):

PR #196, opened 2021-08-29
Rails admin quotes

raccube:
This kinda irks me


pixeldesu:
You are including too many commits here!

PR #204, opened 2021-10-03
Only count active users on about page

### LINKED ISSUE #203: Only count active users on about page
Currently the about page just returns the result of `User.count`

Instead it should not count Users that are:
* not confirmed
* banned
* have no answers/questions

pixeldesu:
Closes #203

### DIFF
--- a/app/controllers/static_controller.rb
+++ b/app/controllers/static_controller.rb
@@ -18,7 +18,7 @@ class StaticController < ApplicationController
 
   def about
     @users = User
-      .where('confirmed_at IS NOT NULL')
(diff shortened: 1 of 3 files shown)


app/controllers/static_controller.rb
```diff
@@ -17,7 +17,18 @@ def index
   end
 
   def about
-
+    @users = User
+      .where('confirmed_at IS NOT NULL')
```
raccube: Think `where.not(confirmed_at: nil)` might work

raccube (approved):

PR #221, opened 2021-12-22
Add accept attribute to file inputs

### LINKED ISSUE #217: Add `accept` attribute to `input[type=file]` fields
Some users have tried to upload avatars/header images in formats we don't support. We can reduce the number of instances of this happening if we add an `accept` attribute to the file upload fields.

I think this will also get iOS to automatically convert HEIF to JPEG.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept

raccube:
Closes #217

### DIFF
--- a/config/justask.yml.example
+++ b/config/justask.yml.example
@@ -60,6 +60,10 @@ hcaptcha:
 # TOTP Drift period in seconds
 otp_drift_period: 30
 
+# This list controls the "accept" attribute on file upload fields
+# This ensures mobile users get an appropriate file picker (one for only images)
+# as well as preventing the upload of videos or formats we don't support
+# including making iOS automatically convert HEIC files to JPEG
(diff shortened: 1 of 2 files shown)


config/justask.yml.example
```diff
@@ -59,3 +59,12 @@ hcaptcha:
 
 # TOTP Drift period in seconds
 otp_drift_period: 30
+
+accepted_image_formats:
```
nilsding, line 67: please add a comment to this describing what it's used for

nilsding (changes requested):

config/justask.yml.example
raccube: a38c1342

nilsding (approved):

pixeldesu (approved):

PR #222, opened 2021-12-22
Add mute rules

### LINKED ISSUE #216: Custom word blacklist/filter for question submissions
one of the things i liked about curiouscat that i've not been able to find on any other similar site, is the ability to have a custom word blacklist/filter that if included in the question, would prevent the question being sent or send the question into the void never ot be seen again instead of into someone's inbox. At most, sites have had a checkbox that would filter questions based on their undisclosed filter system, often based around profanity. It'd be nice to be able to filter out/prevent questions being submitted that mention certain words determined by the user.

raccube:
Closes #216

### DIFF
--- a/app/views/user/edit_mute.haml
+++ b/app/views/user/edit_mute.haml
@@ -1,0 +1,4 @@
+= render 'settings/muted'
+
+- provide(:title, generate_title('Muted words'))
+- parent_layout 'user/settings'
--- a/app/controllers/ajax/mute_rule_controller.rb
+++ b/app/controllers/ajax/mute_rule_controller.rb
@@ -1,0 +1,67 @@
+class Ajax::MuteRuleController < AjaxController
+  def create
+    params.require :muted_phrase
+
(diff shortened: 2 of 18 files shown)


pixeldesu (approved):
LGTM :+1:

app/views/user/edit_mute.haml
```diff
@@ -0,0 +1,4 @@
+= render 'settings/muted', qr_svg: @qr_svg, recovery_code_count: @recovery_code_count
```
nilsding: qr_svg?

nilsding (changes requested):
small stuff, also maybe it makes sense to not use an incremental (by 1) id for these?

app/views/user/edit_mute.haml
raccube: Wow, I blindly the template from `edit_security`.

nilsding (approved):

PR #226, opened 2021-12-25
Port User functionality to TypeScript

### LINKED ISSUE #166: Port User functionality to TypeScript
In an effort to refactor the frontend portion of the site, we want to move from CoffeeScript to TypeScript and in the very same step also drop stuff like jQuery to the bare minimum.

This is the sub-issue of the refactor for the user functionality.

**To Do:**
* [ ] (Un)following user
* [ ] Reporting user

pixeldesu:
_Fixes #166_

To explain some of the changes:

* Technically not in scope, but fits here: More aliases so utilities don't require moving up all the folders but can be imported with just `utilities/` instead.
* Added an `$.on` substitute in vanilla JS, mainly for the use case of global event listeners, which are used plenty in Retrospring, this will make further porting easier.
* The event `initUser` listens on is `DOMContentLoaded` and not `turbolinks:load`. Why? Don't need to set up global event handlers on every turbolinks navigation, once is enough for page scope.
* The header parallax code is gone without a replacement. Looking at my own tests it probably didn't even really work for a while anyway.

To test this PR properly:
* (Un)follow a user
  * ...on a profile
  * ...in the user list on a profile
* Report a user


app/javascript/retrospring/utilities/on.ts
```diff
@@ -0,0 +1,9 @@
+type OnCallbackFunction = (event: Event) => void;
```
raccube, line 1: nit: could be called `EventHandler`

raccube (approved):

PR #229, opened 2021-12-26
Add web app manifest and updated icons for PWA

raccube:


raccube:
This is kind of blocked as we'd need to get an Apple Developer Licence as Safari doesn't provide the `PushService` API that other browsers do. Instead they have their own one which uses APN.

nilsding:
just threw loadsamoney at , so you're no longer blocked on this

nilsding:
maybe this gem could be of interest for this PR? https://github.com/rpush/rpush

raccube:
Good news! Push is supported in iOS 15.4 but behind a feature flag.

raccube:
Narrowing the scope of this PR to just PWA as adding WebPush would inflate this PR drastically.

pixeldesu (approved):

PR #232, opened 2021-12-26
Add manual share links to answerboxes

pixeldesu:
This is mainly there to "circumvent" the issue with automated sharing, offering links to the share widgets of Twitter and Tumblr.

Also moves the general text preparation into own `SocialHelper::` modules, so that it can be reused without duplicating code. Also added tests for all new things!

Also fixes the bug with the OS share menus which only worked on the initial set of answers, that now works on all of them.

### DIFF
--- a/app/helpers/social_helper/tumblr_methods.rb
+++ b/app/helpers/social_helper/tumblr_methods.rb
@@ -1,0 +1,37 @@
+require 'cgi'
+
+module SocialHelper::TumblrMethods
+  include Rails.application.routes.url_helpers
+  
+  def tumblr_title(answer)
+    asker = if answer.question.author_is_anonymous?
+      APP_CONFIG['anonymous_name']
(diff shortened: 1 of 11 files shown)


app/helpers/social_helper/tumblr_methods.rb
```diff
@@ -0,0 +1,35 @@
+require 'cgi'
+
+module SocialHelper::TumblrMethods
+  def tumblr_title(answer)
+    asker = if answer.question.author_is_anonymous?
+      APP_CONFIG['anonymous_name']
+    else
+      answer.question.user.profile.display_name.blank? ? answer.question.user.screen_name : answer.question.user.profile.display_name
```
raccube:
```suggestion
      answer.question.user.profile.safe_name
```

raccube (approved):

PR #238, opened 2021-12-27
Catch common `ShareWorker` exceptions to avoid unnecessary retries

raccube:


app/workers/share_worker.rb
```diff
@@ -11,6 +11,15 @@ def perform(user_id, answer_id, service)
     user_service = User.find(user_id).services.find_by(type: service_type)
 
     user_service.post(Answer.find(answer_id))
+  rescue ActiveRecord::RecordNotFound
+    # The question to be posted was deleted
```
nilsding, line 16: please add some log messages (and maybe some `NewRelic::Agent.notice_error(e)`s) to these so we know how often this happens

nilsding (changes requested):

nilsding (approved):

PR #243, opened 2021-12-28
Port List functionality to TypeScript

### LINKED ISSUE #163: Port List functionality to TypeScript
In an effort to refactor the frontend portion of the site, we want to move from CoffeeScript to TypeScript and in the very same step also drop stuff like jQuery to the bare minimum.

This is the sub-issue of the refactor for the list functionality.

**To Do:**
* [x] Adding new lists
* [x] Adding/removing members of lists
* [x] Removing lists
* [x] Keyboard shortcuts

pixeldesu:
Fixes #163

Should be straightforward to test, just create/delete a list and add/remove a member.

Also fixed a missing translation, that always was here, but never actually set!


app/javascript/retrospring/features/lists/create.ts
```diff
@@ -0,0 +1,34 @@
...
+    },
+    error: (data, status, xhr) => {
+      console.log(data, status, xhr);
+      window['showNotification'](I18n.translate('frontend.error.message'), false);
+    }
+  });
+}
+
+export function createListInputHandler(event: KeyboardEvent): void {
+  if (event.which === 13) {
```
raccube, line 31: A comment of what key this is would be nice

raccube (changes requested):
One small thing.

raccube (approved):

PR #249, opened 2021-12-30
Add Link Proxy

### LINKED ISSUE #110: Add a link proxy to URLs outside of Retrospring
Internally, we reconsidered adding the ability to actually parse links in questions, no matter the origin.

But something we definitely need to add for this to work is adding a "Link Proxy", which basically shows a warning page akin to "You're leaving Retrospring for [full url], click here if you are sure".

**Example:** https://steamcommunity.com/linkfilter/?url=https://retrospring.net

It just can be a simple action in `StaticController` taking the target URL as an argument.

Something up for discussion should be the URL/route for the action? If you have an idea, feel free to propose something.

pixeldesu:
Fixes #110 

Adds formatting to questions, but the only thing that is allowed are links. And if links are not the original host (or a whitelisted one) a link proxy will be shown instead, warning users of potential risk ahead.

To test this locally, don't forget to set the list of allowed hosts in `justask.yml`


raccube (approved):

spec/helpers/markdown_helper_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+
+  describe '#question_markdown' do
+    it 'should link allowed links without the linkfilter' do
+      expect(question_markdown('https://twitter.com/retrospring')).to eq('<p><a href="https://twitter.com/retrospring" target="_blank">https://twitter.com/retrospring</a></p>')
+    end
+
+    it 'should link untrusted links with the linkfilter' do
+      expect(question_markdown('https://rrerr.net')).to eq('<p><a href="/linkfilter?url=https%3A%2F%2Frrerr.net" target="_blank">https://rrerr.net</a></p>')
+    end
+  end
```
nilsding: what's `#question_markdown`'s result on e.g...?

```markdown
**your account has been disabled**, [click here to enable it again](https://evil.example.com)
```

app/services/shared_markers.rb
```diff
@@ -0,0 +1,9 @@
+module SharedMarkers
+  def autolink(link, _link_type)
+    if ALLOWED_HOSTS.include? URI(link).host
+      return "<a href=\"#{link}\" target=\"_blank\">#{link}</a>"
+    end
+
+    "<a href=\"#{linkfilter_path(url: link)}\" target=\"_blank\">#{link}</a>"
```
nilsding: why not use `content_tag` here?

e.g. something like

```ruby
href = if ALLOWED_HOSTS.include?(URI(link).host)
         link
       else
         linkfilter_path(url: link)
       end

content_tag(:a, link, href: href, target: "_blank") # `link` will also be properly HTML escaped
```

I think these links should also be ignored by search engines, so the final content_tag would be e.g.

```ruby
content_tag(:a, link, href: href, target: "_blank", rel: "nofollow")
```

config/initializers/redcarpet.rb
```diff
@@ -25,3 +25,8 @@
   strikethrough: true,
   superscript: false
 }
+
+ALLOWED_HOSTS = [
```
nilsding: the name of this const is too generic.  Perhaps something like `ALLOWED_HOSTS_IN_MARKDOWN` is better here

app/views/static/linkfilter.haml
```diff
@@ -0,0 +1,19 @@
...
+      %p
+        Never enter your passwords or other private information on an untrusted website.
+        = APP_CONFIG['site_name']
+        will only ever ask for your password on a site that is on the domain
+        = APP_CONFIG['hostname']
+
+      %p.font-weight-bold.mb-0 URL:
+      %pre.bg-light.text-dark.p-3.rounded= @link
+
+      %a.btn.btn-primary{ href: @link } I understand the risk, proceed!
```
nilsding: add a `rel: "nofollow"` here so search engines don't count anything passed to the linkfilter

nilsding (changes requested):

app/services/shared_markers.rb
raccube: +1 for the `rel=nofollow`

spec/helpers/markdown_helper_spec.rb
pixeldesu: literally no markup, that's the result


only link and paragraph get processed so there's actual output

nilsding (approved):

PR #254, opened 2021-12-31
Refactor relationships

### LINKED ISSUE #12: Blocking Users
Blocking users you don't like/have personal issues with, the same thing as on every social network.
- X's Profile not visible for user Y
- X's Questions and answers not visible for user Y
- Y's Profile not visible for user X
- Y's Questions and answers not visible for user X

This will cause that these 2 users, except if they end up making new accounts (for stalking), being practically invisible for each other, blocking one user basically shows you have either personal or any issues with them and don't want to interact with them in any way.

raccube:
~~Based on #194, needs to be rebased after that has been merged~~


app/controllers/ajax/relationship_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+require "use_case/relationship/destroy"
+require "errors"
+
+class Ajax::RelationshipController < AjaxController
+  before_action :authenticate_user!
+
+  def create
+    UseCase::Relationship::Create.call(
+      current_user_id: current_user.id,
+      target_user: params[:target_user],
```
pixeldesu: TypeScript is submitting a `screen_name` here, which causes a `param is missing` error on all requests using the UI. Either change this or change it in TypeScript (`retrospring/features/user/action.ts`)

app/javascript/retrospring/features/user/action.ts
```diff
@@ -7,14 +7,15 @@ export function userActionHandler(event: Event): void {
...
-  const targetURL = action === 'follow' ? '/ajax/create_friend' : '/ajax/destroy_friend';
+  const targetURL = action === 'follow' ? '/ajax/create_relationship' : '/ajax/destroy_relationship';
   let success = false;
 
   Rails.ajax({
     url: targetURL,
     type: 'POST',
     data: new URLSearchParams({
-      screen_name: target
+      screen_name: target,
```
pixeldesu, line 17: A `screen_name` is submitted, and a `target_user` is expected in the controller.

app/controllers/ajax/relationship_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def destroy
+    UseCase::Relationship::Destroy.call(
+      current_user_id: current_user.id,
+      target_user: params[:target_user],
+      type: params[:type]
+    )
+    @response[:success] = true
+    @response[:message] = t(".success")
```
pixeldesu: Controller-scoped ones don't work because you didn't provide any, and getting a "translation missing" here would not be desirable.

Use `messages.friend.destroy.okay` instead, or change the key.

app/controllers/ajax/relationship_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  before_action :authenticate_user!
+
+  def create
+    UseCase::Relationship::Create.call(
+      current_user_id: current_user.id,
+      target_user: params[:target_user],
+      type: params[:type]
+    )
+    @response[:success] = true
+    @response[:message] = t(".success")
```
pixeldesu: Controller-scoped ones don't work because you didn't provide any, and getting a "translation missing" here would not be desirable.

Use `messages.friend.create.okay` instead, or change the key.

pixeldesu (changes requested):

pixeldesu (approved):

config/locales/en.yml
```diff
@@ -457,3 +457,24 @@ en:
...
+    param_is_missing: "param is missing"
+
+    forbidden: "This is illegal, you know"
+    blocked: "You have been blocked from performing this request"
+    other_blocked_self: "You have been blocked by this user"
+    asking_other_blocked_self: "You have been blocked from asking this user questions"
+    following_other_blocked_self: "You have been blocked from following this user"
+    self_blocked_other: "You cannot do this while blocking this user"
+    asking_self_blocked_other: "You cannot ask an user who you are currently blocking"
+    following_self_blocked_other: "You cannot follow an user who you are currently blocking"
```
nilsding, line 473: \*notices your new strings\* owo what's this?

config/routes.rb
```diff
@@ -135,12 +135,12 @@
   match '/@:username/a/:id', to: 'answer#show', via: 'get', as: :show_user_answer_alt
   match '/@:username/q/:id', to: 'question#show', via: 'get', as: :show_user_question_alt
   match '/@:username/followers(/p/:page)', to: 'user#followers', via: 'get', as: :show_user_followers_alt, defaults: {page: 1}
-  match '/@:username/friends(/p/:page)', to: 'user#friends', via: 'get', as: :show_user_friends_alt, defaults: {page: 1}
+  match '/@:username/followings(/p/:page)', to: 'user#followings', via: 'get', as: :show_user_followings_alt, defaults: {page: 1}
```
nilsding, line 138: should we make a redirect from the old `/@:username/friends` url to the new one?

db/schema.rb
```diff
@@ -323,6 +323,5 @@
     t.index ["user_id"], name: "index_users_roles_on_user_id"
   end
 
-  add_foreign_key "mute_rules", "users"
```
nilsding: this shouldn't disappear out of nowhere, please run migrations again

lib/use_case/relationship/create.rb
```diff
@@ -0,0 +1,47 @@
...
+          extra:    {
+            target_user: target_user
+          }
+        }
+      end
+
+      private
+
+      def find_source_user
+        return current_user if current_user.is_a?(::User)
```
nilsding: `current_user` can never be a `::User` because of the type restriction

(side note: if the type is extended to also accept `::User` objects, you can pass in the `current_user` object directly to the UseCase call)

lib/use_case/relationship/create.rb
```diff
@@ -0,0 +1,47 @@
...
+
+        ::User.find_by!(screen_name: current_user)
+      rescue ActiveRecord::RecordNotFound
+        raise Errors::UserNotFound
+      end
+
+      def find_target_user
+        return target_user if target_user.is_a?(::User)
```
nilsding, line 39: `target_user` can never be a `::User` because of the type restriction

lib/use_case/relationship/destroy.rb
```diff
@@ -0,0 +1,43 @@
...
+          extra:    {
+            target_user: target_user
+          }
+        }
+      end
+
+      private
+
+      def find_source_user
+        ::User.find_by!(screen_name: current_user)
```
nilsding: ...interestingly enough, in this UseCase `current_user` could be a `::User` object :D 

so you could just:
```suggestion
      def find_source_user
        return current_user if current_user.is_a?(::User)

        ::User.find_by!(screen_name: current_user)
```

lib/use_case/relationship/destroy.rb
```diff
@@ -0,0 +1,43 @@
...
+      rescue ActiveRecord::RecordNotFound
+        raise Errors::UserNotFound
+      end
+
+      def find_target_user
+        ::User.find_by!(screen_name: target_user)
```
nilsding: likewise here

lib/use_case/relationship/create.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+require "use_case/base"
+require "errors"
+
+module UseCase
+  module Relationship
+    class Create < UseCase::Base
+      option :current_user, type: Types::Coercible::String
```
nilsding: I think it's better to call this option `source_user`.

lib/use_case/relationship/destroy.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require "use_case/base"
+require "errors"
+
+module UseCase
+  module Relationship
+    class Destroy < UseCase::Base
+      option :current_user, type: Types::Coercible::String | Types.Instance(User)
```
nilsding: I think it's better to call this option `source_user`.

nilsding (changes requested):

app/controllers/ajax/relationship_controller.rb
```diff
@@ -11,9 +11,9 @@ def create
     params.require :screen_name
 
     UseCase::Relationship::Create.call(
-      current_user: current_user.screen_name,
-      target_user:  params[:screen_name],
-      type:         params[:type]
+      source_user: current_user.screen_name,
```
nilsding: now you can do:
```suggestion
      source_user: current_user,
```

app/controllers/ajax/relationship_controller.rb
```diff
@@ -25,9 +25,9 @@ def create
 
   def destroy
     UseCase::Relationship::Destroy.call(
-      current_user: current_user.screen_name,
-      target_user:  params[:screen_name],
-      type:         params[:type]
+      source_user: current_user.screen_name,
```
nilsding: now you can do:
```suggestion
      source_user: current_user,
```

nilsding (changes requested):

nilsding (approved):

PR #272, opened 2022-01-05
Add the ability to post to services with a tag

raccube:


pixeldesu (changes requested):
Generally, the settings look fine and the posting works well too.

Except for the fact that I can put mentions into the "tag" field and mention users through that, pretty much enabling spam

ex. https://twitter.com/pixeldesu/status/1478767661853167626

pixeldesu (approved):
@-symbols are now stripped from the tag field and HTML5 validation doesn't even allow them anymore, so we should be good now!

PR #278, opened 2022-01-06
Cleanup

nilsding:
Removed some files and replaced the Ruby action with the official one from the Ruby project.


raccube (approved):
(out of scope for this PR) We should probably also cache Node packages

pixeldesu (approved):

PR #279, opened 2022-01-06
Fix incorrect selector in inbox keypress handling

raccube:


pixeldesu:
Superfluous.

The handler is already registered here: https://github.com/Retrospring/retrospring/blob/master/app/javascript/retrospring/features/inbox/entry/index.ts

pixeldesu (approved):

PR #286, opened 2022-01-08
Remove unused static files

pixeldesu:
Conversation from Mumble the past days:
> @raccube: Hey, why does JavaScript have 7.5% in the repository?
> me: Ah, that's surely because of the Webpacker config files!

That only was _halfway_ correct. Before the `npm` and Webpacker days, we included some external JS as vendor files directly into the project. That functionality that required that is either fully removed now or actually has a module required that doesn't need to be present locally.

Also disabled goddamn i18n.js exporting it's main JS file into `public/javascripts/` all the time, we use the node module for that as well.

_You could test this if you want, but these files weren't touched by anything Rails in years anyway, nothing changed._


nilsding:
What does this mean for deployment?  Currently the rake task `i18n:js:export` is run during each deployment, is this still required?

pixeldesu:
> What does this mean for deployment? Currently the rake task `i18n:js:export` is run during each deployment, is this still required?

Yes. `i18n:js:export` does what it says, it exports locales into a JS file.

What was laying in `public/javascripts` however was the runtime (exported by the gem by default), that makes the translation methods available. We never even used that one (we use the version the library provides over NPM). So it can be safely removed.

raccube (approved):

nilsding (approved):

PR #288, opened 2022-01-09
Port announcements functionality to TypeScript

raccube:


app/javascript/packs/application.ts
```diff
@@ -1,6 +1,7 @@
 require('sweetalert/dist/sweetalert.css');
 
 import start from 'retrospring/common';
+import initAnnouncements from 'retrospring/features/announcement/index';
```
pixeldesu:
```suggestion
import initAnnouncements from 'retrospring/features/announcement';
```

You don't need to add `index` to a folder import if there is an `index.{js.ts}` file present.

app/javascript/packs/legacy.coffee
```diff
@@ -50,18 +50,7 @@ _ready = ->
...
-    unless (window.localStorage.getItem("announcement#{aId}"))
-      $(this).toggleClass("d-none")
-
-  $(document).on "click", ".announcement button.close", (evt) ->
-    announcement = event.target.closest(".announcement")
-    aId = announcement.dataset.announcementId
-    window.localStorage.setItem("announcement#{aId}", true)
-
-
+Port
```
pixeldesu:
```suggestion
```

Included by mistake?

app/javascript/retrospring/features/announcement/close.ts
```diff
@@ -0,0 +1,5 @@
+export default function (event: Event): void {
+  const announcement = (event.target as HTMLElement).closest(".announcement") as HTMLDivElement;
+  const aId = announcement.dataset.announcementId;
```
pixeldesu: Variables should have a clear name. Either name it `announcementId` or just `id` (because there aren't multiple we don't need to contextualize it)

pixeldesu (changes requested):

app/javascript/packs/legacy.coffee
raccube: Thought I force pushed this one out

app/javascript/retrospring/features/announcement/index.ts
```diff
@@ -0,0 +1,14 @@
+import registerEvents from 'utilities/registerEvents';
+import closeAnnouncementHandler from './close';
+
+export default (): void => {
+  registerEvents([
+    { type: 'click', target: '.announcement button.close', handler: closeAnnouncementHandler, global: true },
```
pixeldesu:
```suggestion
    { type: 'click', target: document.querySelector('.announcement button.close'), handler: closeAnnouncementHandler },
```

If you are using `turbolinks:load` as an event for a feature, you can't use global events (or, it doesn't work particularly well).

Switching the event to `DOMContentLoaded` or removing `global: true` and query-selecting an element in `target`, like I did in this suggestion.

pixeldesu (changes requested):
I tried creating an announcement and dismissing it, but the handler code didn't trigger. I made suggestions and comments as to why that is, and verified that locally.

app/javascript/retrospring/features/announcement/index.ts
```diff
@@ -0,0 +1,14 @@
+import registerEvents from 'utilities/registerEvents';
+import closeAnnouncementHandler from './close';
+
+export default (): void => {
+  registerEvents([
+    { type: 'click', target: document.querySelector('.announcement button.close'), handler: closeAnnouncementHandler },
+  ]);
+
+  document.querySelectorAll('.announcement').forEach(function (el: HTMLDivElement) {
+    if (!window.localStorage.getItem(el.dataset.announcementId)) {
```
pixeldesu:
```suggestion
    if (!window.localStorage.getItem(`announcement${el.dataset.announcementId}`)) {
```

without this, it queries for a key `1` in Local Storage

pixeldesu (changes requested):

pixeldesu (approved):

PR #295, opened 2022-01-11
Use Webpacker for all assets

### LINKED ISSUE #95: Serve webfont ourselves
Just to not forget about it when we start moving everything to Webpacker...

[Right now](https://github.com/Retrospring/retrospring/blob/1d7430c3f786f42bb5e416eb4596f426d57a19a9/app/assets/stylesheets/application.css.scss#L11) the Lato webfont is served by E-Corp.  We really should host this ourselves.

pixeldesu:
This pretty much moves all SCSS into Webpacker so that it doesn't require asset gems and alike anymore.

Also:
* removed nprogress as it was unused
* moved font-awesome from an asset gem to an npm package
* included Lato font locally, not from Google Fonts (Fixes #95)
* moved the CSS imports in JS to SCSS
* removed all the (now) unused asset gems


raccube (approved):
:shipit:

PR #298, opened 2022-01-12
Upgrade to Rails 6.1

raccube:


config/initializers/assets.rb
```diff
@@ -1,22 +1,12 @@
...
 Rails.application.config.assets.version = '1.0'
 
-# Add additional assets to the asset load path
+# Add additional assets to the asset load path.
 # Rails.application.config.assets.paths << Emoji.images_path
 
 # Precompile additional assets.
-# application.js, application.css, and all non-JS/CSS in app/assets folder are already added.
-Rails.application.config.assets.precompile += %w( moderation.js )
-Rails.application.config.assets.precompile += %w( i18n.js )
```
nilsding, line 22: this is not required anymore because of webpacker, right?

db/schema.rb
```diff
@@ -180,7 +180,7 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.index ["name", "resource_type", "resource_id"], name: "index_roles_on_name_and_resource_type_and_resource_id"
-    t.index ["resource_type", "resource_id"], name: "index_roles_on_resource_type_and_resource_id"
+    t.index ["resource_type", "resource_id"], name: "index_roles_on_resource"
```
nilsding, line 183: why did this index name change?

spec/helpers/bootstrap_helper_spec.rb
```diff
@@ -88,9 +88,9 @@
 
   describe "#time_tooltip" do
     it 'should return a tooltip with proper time values' do
-      Timecop.freeze(Time.local(1984)) do
+      Timecop.freeze(Time.utc(1984)) do
```
nilsding, line 91: literally 1984

config/initializers/assets.rb
raccube: If I'm not mistaken we still use the asset pipeline for i18n. I'm also fairly certain rails admin still uses it

spec/helpers/bootstrap_helper_spec.rb
raccube: wortwörtlich 1984

config/initializers/assets.rb
pixeldesu: Yes:
* `moderation.js` as an entrypoint doesn't exist anymore
* `i18n.js` is not used as an exported asset anymore either

db/schema.rb
raccube: Just cleared out my db and rerun migrations; this is how it generated.

nilsding: weird.

nilsding (approved):

pixeldesu (approved):
Tested through the application, everything still seems to work fine! LGTM 👍

PR #305, opened 2022-01-15
Set some stricter Rubocop rules

pixeldesu:
We don't really check our code with Rubocop at the moment, and we should change that.

This configuration is based on a project we wrote in Rails before and includes a sane (but strict) set of rules to follow.

After we merge this, we should do one of the following:
* Enable some kind of SaaS GitHub PR integration (like HoundCI or Codacy) that checks new code for style violations
  * ...with the negative that they most likely don't support arbitrary imports (like `rt_rubocop_defaults`)
* Add a GitHub Actions step that runs `bundle exec rubocop`
  * ...which we need to set to `continue-on-error` until we fixed all violations
* Setup https://github.com/reviewdog/action-rubocop to run on PRs
  * ...which requires a new GitHub user of which we use the token to pass to the action to gives us comments on PRs

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,7 +2,7 @@ require:
   - rt_rubocop_defaults
   - rubocop-rails
 AllCops:
-  TargetRubyVersion: 2.5
+  TargetRubyVersion: 2.7
   Exclude:
     - 'config/**/*'
     - 'vendor/**/*'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,8 +1,8 @@
-require: 
-  - rt_rubocop_defaults
+require:
   - rubocop-rails
 AllCops:
-  TargetRubyVersion: 2.7
+  DisplayCopNames: true
+  CacheRootDirectory: '.git'
   Exclude:
     - 'config/**/*'
     - 'vendor/**/*'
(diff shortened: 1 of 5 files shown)


.rubocop.yml
```diff
@@ -1,23 +1,43 @@
+require: 
+  - rt_rubocop_defaults
+  - rubocop-rails
 AllCops:
-  NewCops: enable
+  TargetRubyVersion: 2.5
```
raccube: Could change this to 2.7 as that's what's currently running in prod.

nilsding: would also suggest to just bump this to 2.7, especially since I already made use of some 2.7 feature in #294 (numbered parameters, the`_1` thingies)

```suggestion
  TargetRubyVersion: 2.7
```

nilsding (changes requested):
please change the TargetRubyVersion

nilsding (approved):

Gemfile
```diff
@@ -94,6 +94,9 @@ group :development, :test do
   gem 'timecop'
   gem 'rails-controller-testing'
   gem 'haml_lint', require: false
+  gem 'rubocop', '~> 1.22', '>= 1.22.1'
+  gem 'rubocop-rails', '~> 2.13', '>= 2.13.1'
+  gem 'rt_rubocop_defaults', '~> 2.3', '>= 2.3.1'
```
nilsding: Imposter sus

raccube: Good dog

pixeldesu:
Looks like Hound is not using the configuration from the PR (which kinda makes sense), so we'll just keep getting default suggestions until we merge this.

_Yeet._

raccube (approved):

PR #308, opened 2022-01-16
Fix code style in ApplicationHelpers

pixeldesu:
Now that we have adjusted Rubocop styles, we can slowly adjust the style across all files to be clean, this starts with `ApplicationHelper::`

**Why did the tests change?**  
One of the issues presented by Rubocop was the usage of `.html_safe` in two helper methods. I replaced the `.join("\n").html_safe` calls with `safe_join(..., "\n")`, which produces different HTML output (the closing `/>` in the meta tags turned to `>` which is still valid HTML5, for empty self-closing tags)

Addendum: Actually, the difference in tag markup probably comes from using `tag.meta` instead of `tag(:meta...`


pixeldesu:
Finally, finally appeased the dog overlords.

raccube:
But you did not appease the umbrella overlords.

app/helpers/application_helper/title_methods.rb
```diff
@@ -1,28 +1,26 @@
...
 
-  def generate_title(name, junction = nil, content = nil, s = false)
-    if s
-      if name[-1].downcase != "s"
-        name = name + "'s"
-      else
-        name = name + "'"
-      end
+  def generate_title(name, junction = nil, content = nil, possesive = false)
+    if possesive
```
nilsding: typo:
```suggestion
  def generate_title(name, junction = nil, content = nil, possessive = false)
    if possessive
```

nilsding (approved):

PR #314, opened 2022-01-18
Add Brewfile

raccube:


README.md
```diff
@@ -7,9 +7,14 @@ development.
 
 ## Installation
 
-You can find all the installation instructions needed for a local/production
-setup of Retrospring in the
-[Wiki](https://github.com/Retrospring/retrospring/wiki/Setup).
```
pixeldesu: Suddenly, no other operating system than MacOS exists anymore.

Also, why not add this to the wiki, where the other installation instructions already are?

pixeldesu (changes requested):

README.md
raccube: I guess that makes sense, although the installation instructions currently in the wiki are very outdated.

pixeldesu: Well, we have two measures then, either:
* ...add updated installation instructions here
* ...or in the wiki

Also, your installation instructions just mention how to install the dependencies, but not how to run it in the end (nor the fact what needs to be run or configured in what way).

Considering there's quite a few steps involved, the wiki is probably a better place, rather than blowing up the README with this.

pixeldesu: Done the deed, updated the setup instructions:
https://github.com/Retrospring/retrospring/wiki/Setup#installation

pixeldesu (approved):

PR #315, opened 2022-01-22
Integrate feedback forms with Canny

### LINKED ISSUE #90: Add a form on the site for feedback/feature requests/bug reports.
Currently the only way to submit any feedback for Retrospring is to create an issue in this issue tracker. This requires the user to have a GitHub account, likely deterring users from submitting their feedback or suggestions.

We have a few options for how we could tackle this:

- Creating a form which simply opens a GitHub issue (could use @Kaisrrerr for this)
- The above but with some kind of admin approval before opening an issue on GitHub
- Repurposing the current moderation system to create a "Feedback Hub" of sorts where users can vote on issues

pixeldesu:
This integrates feedback and bug report forms with [Canny](https://canny.io)

Wasn't sure where to put the links so they are at least somewhat visible (but not a main feature), so I put them into the profile/user dropdown:
image

**Prerequisites**
* Get Canny credentials:
  * SSO token: https://developers.canny.io/install/widget/sso
  * Board tokens: https://developers.canny.io/install/widget/web
  * And put them into `config/justask.yml`

**Testing**
* Try accessing `/feedback/bugs`
  * This should redirect to `/feedback/consent`
* After consenting, try accessing `/feedback/consent`
  * This should redirect to `/feedback/bugs`
* On one of the form pages, check one of the script tags after the feedback form, including an `ssoToken`
* Create an example post in one of the boards
  * If using an admin account (determined by the used email) this will be directly visible
  * Otherwise the screen name of the user and the profile picture (which won't work since the URL is only local) should be set
* If the `canny` section of the comments is commented out, visiting a feedback path should redirect to `root_path`


app/views/navigation/mobile/_profile.haml
```diff
@@ -24,8 +24,17 @@
...
     .dropdown-divider
-  = button_to destroy_user_session_path, method: 'delete', class: 'dropdown-item' do
+  - if APP_CONFIG["canny"]
+    %h6.dropdown-header.d-none.d-sm-block Feedback
+    %a.dropdown-item{ href: feedback_features_path }
+      %i.fa.fa-fw.fa-flask
+      Feature Requests
+    %a.dropdown-item{ href: feedback_bugs_path }
+      %i.fa.fa-fw.fa-bug
+      Bugs
```
raccube:
```suggestion
    %a.dropdown-item{ href: feedback_bugs_path }
      %i.fa.fa-fw.fa-bug
      Bugs
    %a.dropdown-item{ href: feedback_features_path }
      %i.fa.fa-fw.fa-flask
      Feature Requests
```

nit: to line up with the tabs also puts these in alphabetical order

.hound.yml
```diff
@@ -8,6 +8,7 @@ jshint:
   enabled: false
 
 haml:
+  enabled: false
```
raccube, line 11: thought: could remove the "Lint Haml Templates" step from our CI build and do that here instead?

raccube (changes requested):

.hound.yml
pixeldesu: I removed the HAML lint step because Hound does it as well, which is better for the review/PR approach.

But haml-lint infers RuboCop cops (locally) and that makes sense for stuff like quote rules and such, and Hound can't resolve these either, so either we go for a single-quote style there and disable Cop stuff in HAML, or we just don't lint with Hound here as well.

app/views/navigation/mobile/_profile.haml
pixeldesu: If you adjust the order here, then also do it in the desktop navigation

raccube (approved):

app/views/navigation/mobile/_profile.haml
```diff
@@ -24,8 +24,17 @@
   - if current_user.mod?
     %a.dropdown-item{ href: moderation_path }
       %i.fa.fa-fw.fa-gavel
-      = t('views.navigation.moderation')
+      = t("views.navigation.moderation")
     .dropdown-divider
-  = button_to destroy_user_session_path, method: 'delete', class: 'dropdown-item' do
+  - if APP_CONFIG["canny"]
+    %h6.dropdown-header.d-none.d-sm-block Feedback
```
nilsding, line 30: I hope having all these nav entries will still look fine on smaller mobile screens... :P

config/routes.rb
```diff
@@ -144,5 +144,9 @@
   match '/:username/lists(/p/:page)', to: 'user#lists', via: 'get', as: :show_user_lists, defaults: {page: 1}
   match '/:username/questions(/p/:page)', to: 'user#questions', via: 'get', as: :show_user_questions, defaults: {page: 1}
 
+  match '/feedback/consent', to: 'feedback#consent', via: ['get', 'post'], as: 'feedback_consent'
```
nilsding: I think using the same action for both GET and POST is a bad idea.

Send someone a link to `/feedback/consent?consent=true` and watch them consent in 0.0000005 seconds flat ...

nilsding: Please add `feedback` to the `forbidden_screen_names` list.

spec/helpers/feedback_helper_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+                   sso:           "sso",
+                   feature_board: "feature",
+                   bug_board:     "bug"
+                 }
+               })
+  end
+
+  describe "#canny_token" do
+    context "user signed in" do
+      let(:user) { FactoryBot.create(:user, id: 10, screen_name: "canned_laughter", email: "can@do.com") }
```
nilsding, line 21: I cann't believe this.  🤣

config/justask.yml.example
```diff
@@ -75,3 +75,9 @@ allowed_hosts_in_markdown:
 
 # Sentry connection string
 sentry_dsn: ''
+
+# Canny feedback forms
+# canny:
+#  sso: "CANNY_SSO_TOKEN_HERE"
+#  feature_board: "CANNY_FEATURE_BOARD_TOKEN"
+#  bug_board: "CANNY_BUGS_BOARD_TOKEN"
```
nilsding: Final newline missing

config/routes.rb
```diff
@@ -144,5 +144,9 @@
...
+  match '/feedback/bugs(/*any)', to: 'feedback#bugs', via: 'get', as: 'feedback_bugs'
+  match '/feedback/feature-requests(/*any)', to: 'feedback#features', via: 'get', as: 'feedback_features'
```
nilsding: just an ask: why kebab-case instead of snake_case like the other routes?

app/controllers/feedback_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  before_action :feature_enabled?
+  before_action :canny_consent_given?, only: %w[features bugs]
+
+  def consent
+    redirect_to feedback_bugs_path if current_user.has_role? :canny_consent
+
+    return unless params[:consent] == "true"
+
+    current_user.add_role :canny_consent
+    redirect_to feedback_bugs_path
```
nilsding, line 16: so if a user clicked on `Feature requests` and they didn't consent yet they'll end up at the `Bugs` page?

nilsding (changes requested):

app/views/navigation/mobile/_profile.haml
pixeldesu: Following screenshot is an iPhone SE viewport and the red-boxed area isn't visible to most users, so we should be fine!

J3yiDbCxSE

app/controllers/feedback_controller.rb
pixeldesu: Yes.

I'm a bit conflicted on _how much_ of an issue this ends up being, since the tabs at the top are relatively large.

image

Otherwise it shouldn't be that hard to implement that.

pixeldesu: Will open an issue after this is merged because a simple gate to prevent external URL redirects is only added in Rails 7 🤦

nilsding (approved):

PR #318, opened 2022-01-23
Remove all non-english locales and disable locale switching

pixeldesu:
This is in preparation of the locale refactor, basically netting us time to slowly refactor feature-by-feature.

**Testing:**
* Try setting a locale...
  * ...with the `hl` cookie
  * ...with the `hl` search parameter
  * ...the `locale` field in the `User` model
* All of the above shouldn't change the site being displayed in English


app/views/shared/_locales.haml
```diff
@@ -1,4 +1,4 @@
-.container.d-print-none
+.container.d-print-none.d-none
```
raccube: Why not just not include this partial?

app/controllers/application_controller.rb
```diff
@@ -11,25 +11,25 @@ class ApplicationController < ActionController::Base
...
+    #   else
+    #     I18n.locale = 'en'
+    #   end
+    # else
+    #   I18n.locale = params[:hl]
+    #   if current_user.present?
+    #     current_user.locale = I18n.locale
+    #     current_user.save!
+    #   end
+    # end
```
raccube: Not really a fan of commenting out code.

raccube (approved):

PR #319, opened 2022-01-24
Add translations for Announcements

pixeldesu:
Considering this was new-ish and I specifically gave the order not to localize the strings when implementing, this is now done.

Adding locales for `Announcement`/`AnnouncementController` using the Rails scoping features, akin to

```yml
en:
  controller:
    action:
      message: "Test"
```

So that following works

```rb
def action
  t(".message")
end
```

----

**Testing:**
* Check the views in `/admin/announcements` for proper text/translation display


config/locales/activerecord.en.yml
```diff
@@ -0,0 +1,13 @@
...
+      announcement:
+        one: "Announcement"
+        other: "Announcements"
+    attributes:
+      announcement:
+        content: "Content"
+        link_href: "Link URL"
+        link_text: "Link text"
+        starts_at: "Start time"
+        ends_at: "End time"
```
nilsding: Final newline missing.

nilsding (changes requested):

config/locales/en.yml
```diff
@@ -457,6 +457,27 @@ en:
           invalid_code: "The code you entered was invalid."
         setup:
           success: "Two factor authentication has been enabled for your account."
+  announcement:
+    index:
+      title: :activerecord.models.announcement.other
+      confirm: "Are you sure you want to delete this announcement?"
+      new: "Add Announcement"
```
raccube, line 464:
```suggestion
      new: "Make an Announcement"
```

raccube (changes requested):

config/locales/en.yml
raccube: 🥚

pixeldesu: 🛑

raccube (approved):

nilsding (approved):

nilsding:
# **🐾🐾🐾🐾🐾**

PR #323, opened 2022-01-28
Add titles for feedback pages

raccube:


pixeldesu (changes requested):
Please add translations for the titles.

Use the controller/action scope and give each a `.title` to use.

I'll take care of the rest of Feedback in another PR

pixeldesu (approved):

PR #330, opened 2022-02-04
Use `dependent: :destroy_async` instead of `DeletionWorker`

raccube:


nilsding:
I'm curious about the jobs this enqueues.

pixeldesu (approved):
Tested it, seems to work fine!

I'm a bit curious about something: `resource.destroy` immediately removes the User, and the attached records are deleted asynchronously. So, would it be possible that someone races this on their feed?

e.g.
* Someone deletes themselves
* Someone reloads their timeline
* `User not found for answer xyz`

Since the user is already gone, but an answer is only deleted after the fact.

In general I think we can only really verify if this happens using "real world data" and people actually, eventually running into this (if it's actually possible). Just something I got to think about.

PR #371, opened 2022-04-08
Increase in-feed answer display length

pixeldesu:
Questions can be longer than answers in the feed. This increases answer display length to 640 characters (128 characters more than the 512 for questions)

This is mainly a "quick-fix" to show more content until we refactor the unfolding of content in general.


nilsding (approved):
N🍦

raccube (approved):

PR #376, opened 2022-04-12
Add CodeQL workflow

pixeldesu:
_Let's see what GitHub has to say about our code!_


raccube (approved):
Should we handle the existing failures as part of this PR?

pixeldesu:
@raccube The JS part is relatively easy, I'm just going to wrap the input with `encodeURIComponent`, which should fix it.

I'm not entirely sure about the ServiceController token validation skip, I'm already setting up my project again to test if removing the line still works.

raccube:
Tested locally, works fine.

PR #378, opened 2022-04-14
Configure better_errors to work in Docker environments

pixeldesu:
It's been so long since I used `better_errors`, I want to have it back.


raccube (approved):
🐳

PR #383, opened 2022-04-16
Track version in application

pixeldesu:
This PR implements a module `Retrospring::Version` allowing us to track/display the current version of Retrospring in the application itself, and not just on GitHub.

* The version is sent to Sentry  
  image
* The version is visible on hovering the site title in the footer  
  image

In preparation for forks/custom versions of Retrospring I also added a `suffix` that can be set (like `x.y.zsuffix` in Mastodon)


nilsding (approved):

spec/lib/version_spec.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require "version"
+
+RSpec.describe Retrospring::Version do
+  before(:each) do
+    allow(Retrospring::Version).to receive(:year) { 1984 }
```
raccube, line 7: Literally 1984

raccube (approved):
Means we will need to commit version number bumps but otherwise 👍

pixeldesu:
Yeah, we just need to remember doing thst from now on!

PR #384, opened 2022-04-17
Implement new about page layout

pixeldesu:
The about page was a bit of a problem child of mine for a while, and with the new wave of people interested in Retrospring, and probably even for self-hosting, this page was up for a rework next.

**Screenshot:**
firefox_PiGv6n6uXZ

**What changed:**
* Removed the **Links** and **Open Source** section from the about page
* Removed the FAQ page and links to it (it was viewed so infrequently, the controller action is not listed for Sentry)
* Added configurable about text (in Markdown), which can be adjusted in `justask.yml`
* Adjusted the styling of statistics a bit
* Added locales for the new about page and removed the old ones
* Removed the requirement for users to have asked a question to be included in the user count (and adjusted the test case)

**Testing:**
* Take a look and give feedback


app/views/static/about.haml
```diff
@@ -1,67 +1,31 @@
...
-            %h4.entry__description= t('views.general.smile').pluralize(@smiles)
-        .col-md-3.col-sm-12.col-xs-12
-          .entry.entry--users
-            .entry__value#follower-count= @users
-            %h4.entry__description= t('views.general.user').pluralize(@users)
+  .row
+    .col-sm-8
+      .card.card--fullheight
+        .card-body
+          - if APP_CONFIG["about"].present?
```
nilsding, line 10: I’m sure we’ll forget to set this before the next deploy ;-)

nilsding (approved):

app/views/static/about.haml
pixeldesu: It'll just show

> The site owner decided not to describe this Retrospring instance, how mysterious!

For a while then :trollface:

raccube (approved):

PR #392, opened 2022-06-09
Implement User Blocking

### LINKED ISSUE #12: Blocking Users

raccube:
Closes #12 

Left to do:

- [x] Tests
- [x] Error messaging
- [ ] ~Include blocks in data export~ (will be done in a seperate PR as we don't export relationships at all yet)

### DIFF
--- a/app/controllers/ajax/answer_controller.rb
+++ b/app/controllers/ajax/answer_controller.rb
@@ -32,20 +32,11 @@ class Ajax::AnswerController < AjaxController
       return
     end
 
-    answer = nil
-
-    begin
-      answer = if inbox
-                 inbox_entry.answer params[:answer], current_user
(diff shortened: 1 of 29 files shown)


pixeldesu (approved):
👍 Tested things work fine and as expected!

The error message on trying to smile an answer seems wrong tho
image

PR #396, opened 2022-06-14
Implement blocking for anonymous users

raccube:


pixeldesu (changes requested):
The `AnonymousBlock` model is missing a ` validates_uniqueness_of` check.

Additionally, when blocking two questions from the same user, on the second one, there should be an exception (if the error is related to uniqueness) but in the frontend it should be shown as successfully having blocked that user.

This should be done to prevent users from guessing who someone is by blocking every anonymous user.

----

Aside of that, it works fine! 👍

raccube:
The only problem I see with that is this creates a situation where a user can block an anon and have them not be added to the list of blocked anons.

pixeldesu:
Ah, the check is already just an `.any?` on any filter being set, so that should actually be fine then.

pixeldesu (approved):

app/controllers/ajax/question_controller.rb
```diff
@@ -80,11 +83,10 @@ def create
         return
       end
 
-      # rubocop:disable Style/IfUnlessModifier
-      unless MuteRule.where(user: target_user).any? { |rule| rule.applies_to? question }
+      unless MuteRule.where(user: target_user).any? { |rule| rule.applies_to? question } ||
+          (author_is_anonymous && AnonymousBlock.where(identifier: AnonymousBlock.get_identifier(request.ip)).any?)
```
nilsding: won't this affect anon-blocks site-wide?  i.e., user A blocks anon X, which results in anon X not being able to ask user B anything

better would maybe something along the lines of:
```ruby
target_user.anonymous_blocks.find(identifier: AnonymousBlock.get_identifier(request.ip))
```
(didn't test this, but you'll get the idea)

pixeldesu: I was about to come back here and actually just mention that.

nilsding (changes requested):
one thing :-)

app/controllers/ajax/question_controller.rb
raccube: 6ac8504

db/schema.rb
```diff
@@ -39,6 +50,18 @@
...
+    t.bigint "user_id", null: false
+    t.bigint "parent_id", null: false
+    t.string "parent_type", null: false
+    t.text "content"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+    t.index ["parent_id", "parent_type"], name: "index_appendables_on_parent_id_and_parent_type"
+    t.index ["user_id", "created_at"], name: "index_appendables_on_user_id_and_created_at"
+  end
+
```
nilsding: ?

nilsding (approved):

db/schema.rb
raccube: Oops, that shouldn't be there.

PR #409, opened 2022-06-23
Add method of unmasking anons

### LINKED ISSUE #369: Add mode to unmask anonymous users for moderators
Sometimes going over user profiles/feeds to check for anonymous content is kinda tedious because figuring out if someone is without an account or just set as anonymous requires checking the question using Rails Admin.

Add the ability for moderators to enable (and disable) a mode that shows user information where available.

This can be accomplished using a cookie (so, we don't always need to append something to the URL), and should have an entry in the profile dropdown for easy access.

On the frontend side, this most likely only requires a change here: https://github.com/Retrospring/retrospring/blob/main/app/helpers/user_helper.rb (I don't think it's necessary we show the avatar for unmasked users, it helps distinguishing them from regularly asked questions)

raccube:


app/helpers/user_helper.rb
```diff
@@ -1,23 +1,47 @@
...
+  end
+
+  def should_unmask?(anonymous_identifier)
+    moderation_view? && anonymous_identifier.present?
+  end
+
+  def unmask(user, context_user, anonymous_identifier)
+    return profile_link(user) if user.present?
+
+    "<abbr title='#{anonymous_identifier}'>#{anonymous_name(context_user)}</abbr>"
```
pixeldesu: Please don't return raw HTML, always use `content_tag`

pixeldesu (changes requested):

app/helpers/user_helper.rb
raccube: 863e49a

pixeldesu (approved):

PR #418, opened 2022-06-26
Fix various issues relating to user bans

raccube:
- Fixes ban reasons being set to the screen name of the user being banned
- Fixes banned users unable to logged in (Sentry: `RETROSPRING-3K`)
- Adds tests for attempting to log in while banned
- Moves creation and invalidation of user bans into `User#ban` and `User#unban`
- Removes legacy ban-related fields from the `users` table
- Updates the user count shown on the About to account for the amount of banned users

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -234,9 +234,9 @@ class User < ApplicationRecord
   end
 
   def unban
-    UserBan.current.where(user_id: self.id).update_all(
+    bans.current.update(
       # -1s to account for flakyness with timings in tests
-      expires_at: DateTime.now - 1.second
+      expires_at: DateTime.now.utc - 1.second
     )
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -229,6 +229,10 @@ class User < ApplicationRecord
   end
   # endregion
 
+  def permanently_banned?
+    bans.current.first&.permanent? || false
+  end
+
   def banned?
     self.bans.current.count > 0
   end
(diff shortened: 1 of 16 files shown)


pixeldesu:
When opening user profiles:
```
NoMethodError at /testdesu

undefined method `permanent?' for #<User:0x00007f7658aecdd0>
Did you mean?  permanently_banned
```

app/models/user.rb
```diff
@@ -1,3 +1,8 @@
+# frozen_string_literal: true
+
+require "use_case/user/ban"
```
nilsding: unnecessary require

nilsding (approved):

nilsding (changes requested):
small things

app/models/user.rb
raccube: c724b6e

pixeldesu (approved):

app/models/user.rb
```diff
@@ -1,8 +1,11 @@
+# frozen_string_literal: true
+
 class User < ApplicationRecord
   include User::Relationship
   include User::Relationship::Follow
   include User::Relationship::Block
   include User::AnswerMethods
+  include User::BanMethods
```
nilsding, line 8: hmm.. seeing these `has_many :whatevers` down there makes me think if it would make sense to move them to the `User::WhateverMethods` module (might need to wrap them inside a `included do .. end` block), and renaming these modules to `User::Whatevers`.  But this is nothing for this PR ;-)

nilsding (approved):

PR #422, opened 2022-06-27
Set autocomplete and input mode attributes for auth related fields

raccube:


pixeldesu (approved):
Can't really test this with a mobile device locally but it looks fine code-wise to me 👍

PR #428, opened 2022-07-02
Move export actions into `Settings::ExportController`

pixeldesu:
More refactoring methods out of `UserController`

**Testing:**
* Make sure everything around exporting still works, text looks fine etc.

I fixed an issue with `user_stub` in the exporter, but I can't test the export locally because of HTTParty trying to access whatever as a host.

### DIFF
--- a/app/controllers/settings/export_controller.rb
+++ b/app/controllers/settings/export_controller.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+class Settings::ExportController < ApplicationController
+  before_action :authenticate_user!
+
+  def index
+    if current_user.export_processing
+      flash[:info] = t(".info")
+    end
+  end
+
+  def create
+    if current_user.can_export?
+      ExportWorker.perform_async(current_user.id)
+      flash[:success] = t(".success")
+    else
+      flash[:error] = t(".error")
+    end
+
+    redirect_to settings_export_path
+  end
+end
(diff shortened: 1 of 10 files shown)


app/controllers/settings/export_controller.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class Settings::ExportController < ApplicationController
+  before_action :authenticate_user!
+
+  def index
+    if current_user.export_processing
```
pixeldesu: Huh, I'm not sure if these reports are correct here?

raccube (approved):

app/controllers/settings/export_controller.rb
nilsding: they are correct here, but what i would rather do instead is change this to e.g.

```suggestion
    flash[:info] = t(".info") if current_user.export_processing
```

(and remove the next line)

spec/controllers/settings/export_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe Settings::ExportController, type: :controller do
+  describe "#index" do
```
nilsding, line 6: might as well want to add some basic tests for `#create`

nilsding (approved):

nilsding (approved):

PR #433, opened 2022-07-02
Move 2FA settings actions into `OtpAuthenticationController`

pixeldesu:
Refactooooooring. Also prepares us to cleanly introduce more two-factor options without having to add all of those into the same controller again.

**Testing:**
* Make sure (re)setting and removing 2FA still works fine, locales look good, etc.


app/views/tabs/_settings.haml
```diff
@@ -3,7 +3,7 @@
     = list_group_item t(".account"), edit_user_registration_path
     = list_group_item t(".profile"), settings_profile_path
     = list_group_item t(".privacy"), settings_privacy_path
-    = list_group_item t(".security"), edit_user_security_path
+    = list_group_item t(".security"),settings_two_factor_authentication_otp_authentication_path
```
nilsding:
```suggestion
    = list_group_item t(".security"), settings_two_factor_authentication_otp_authentication_path
```

nilsding (approved):

raccube (changes requested):
image

The translation keys in the `recovery_keys` view are incorrect

raccube (approved):
Need to open another PR to fix the navbar appearing in print

PR #439, opened 2022-07-04
Configure RailsAdmin to use webpacker

raccube:


pixeldesu:
Is there a specific reason that makes this beneficial if Webpacker is deprecated and I'm already in the process of thinking of removing it?

raccube:
After running `assets:precompile` for production locally, I was unable to get any assets to build until I did this.

raccube:
RailsAdmin won't support cssbundling/jsbundling until 3.1 (which is in beta)

app/javascript/packs/rails_admin.js
```diff
@@ -0,0 +1,2 @@
+import "rails_admin/src/rails_admin/base";
+import "../stylesheets/rails_admin.scss";
```
pixeldesu: I'd preferably have this entrypoint as a `.ts` file, if that works. Contents don't need to be modified for that!

pixeldesu (changes requested):

app/javascript/packs/rails_admin.js
raccube: 555f47d

pixeldesu (approved):
👍 LGTM

PR #445, opened 2022-07-06
Add translations for AJAX controllers

pixeldesu:
Mostly just moves locales into the new `controllers.en.yml` file with proper namespacing, plus some adjustments here and there.

There's a lot of improvements to be made here, but these come from logic/validation/guarding and probably will get rid of a bunch of the locales made here, this is just an initial step of moving all the things into proper places.

**Testing:**
* Make sure all AJAX actions return proper messages

b3876017-7b6a-45b9-88e0-6db0053de207_1 9f94b6df6df1fe784c665df21a5a619f

### DIFF
--- a/app/controllers/ajax/answer_controller.rb
+++ b/app/controllers/ajax/answer_controller.rb
@@ -12,7 +12,7 @@ class Ajax::AnswerController < AjaxController
 
       unless current_user == inbox_entry.user
         @response[:status] = :fail
-        @response[:message] = I18n.t('messages.answer.create.fail')
+        @response[:message] = t(".error")
         return
       end
     else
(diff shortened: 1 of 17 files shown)


app/controllers/ajax/answer_controller.rb
```diff
@@ -12,26 +12,19 @@ def create
...
         @response[:status] = :privacy_stronk
-        @response[:message] = I18n.t('messages.answer.create.privacy_stronk')
+        @response[:message] = t(".privacy")
         return
       end
     end
 
-    # this should never trigger because empty params throw ParameterMissing
-    unless params[:answer].length > 0
-      @response[:status] = :peter_dinklage
```
nilsding, line 30: :(

nilsding (approved):

app/controllers/ajax/report_controller.rb
```diff
@@ -31,14 +32,14 @@ def create
...
-      @response[:message] = I18n.t('messages.report.create.not_found', parameter: params[:type])
+      @response[:message] = t(".notfound", parameter: params[:type])
       return
     end
 
     current_user.report object, params[:reason]
 
     @response[:status] = :okay
-    @response[:message] = I18n.t('messages.report.create.okay', parameter: params[:type])
+    @response[:message] = t(".success", parameter: params[:type])
```
raccube:
```suggestion
    @response[:message] = t(".success", parameter: params[:type].titleize)
```

raccube (changes requested):

raccube (approved):

PR #446, opened 2022-07-06
Merge `author_name` with `author_identifier`

raccube:


pixeldesu (approved):
Tested generated questions and the export and it works fine

nilsding:
but https://github.com/Retrospring/retrospring/blob/main/Rakefile#L419 is still referencing author_name :(

PR #453, opened 2022-07-08
Fix notifications being orphaned on parent being deleted

raccube:
For some reason when adding appendables, I didn't think that I could use `notification.target.type` and did something really hacky instead. Turns out this breaks our `Notification#denotify` method

[You Fools! You've messed with the natural order!](https://www.youtube.com/watch?v=sRLlYXKPMUM)

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -31,7 +31,7 @@ class Notification < ApplicationRecord
       notif_type = target.notification_type
       return nil unless notif_type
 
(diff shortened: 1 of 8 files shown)


nilsding (changes requested):
please please please PLEASE add a test for this to ensure we never break this again :D

nilsding (approved):

pixeldesu (approved):
Tested a bunch of notification cases and deleted parent entries etc., works fine!

PR #454, opened 2022-07-09
Remove unneeded Rake tasks

raccube:


nilsding (approved):

pixeldesu:
@nilsding we can actually remove all of the `fix_` tasks, can't we? Pretty sure they were made for weird edge cases or things not properly covered with `dependent`s and having to fix up the database afterwards.

I think we never ran those again after 2015.

nilsding:
we indeed don't, since then we've learnt how to properly use `dependent: :destroy` :P 

removed these tasks and fixed another one

PR #509, opened 2022-07-09
Data exporter improvements

raccube:
Things to test:
- Make sure you can export
- Check the contents of the archive


pixeldesu (approved):
👍 LGTM

Note `/tmp/rs_export` needs to be present on the server, otherwise you just get empty exports. That happened to me and I was confused until I saw the actual error in the Sidekiq log 🥴

PR #519, opened 2022-07-14
Fix issues with account deletion

### LINKED ISSUE #516: Can't delete account
Every time I try to delete my account, I receive the 500 Internal Server Error page. I've tried to do it from both the mobile web app and on my computer. Any way this can be fixed? 

(Also, just want to say I don't have any issue with the website itself; it's great and serves it's purpose, but I'm being harassed by a very determined person across multiple platforms, so I'm trying to limit the ways they can try to communicate with me in hopes that they eventually get bored. Keep up the good work!)

raccube:
Closes #516

### DIFF
--- a/app/views/settings/blocks/index.haml
+++ b/app/views/settings/blocks/index.haml
@@ -25,7 +25,10 @@
         %li.list-group-item
           .d-flex
             %div
-              %p.mb-0= block.question.content
+              - if block.question.present?
+                %p.mb-0= block.question.content
+              - else
+                %p.mb-0.text-muted.font-italic= t('.deleted_question')
               %p.text-muted.mb-0= t(".blocked", time: time_ago_in_words(block.created_at))
             .ml-auto.d-inline-flex
               %button.btn.btn-default.align-self-center{ data: { action: "anon-unblock", target: block.id } }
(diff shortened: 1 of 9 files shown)


app/views/settings/blocks/index.haml
```diff
@@ -25,7 +25,10 @@
         %li.list-group-item
           .d-flex
             %div
-              %p.mb-0= block.question.content
+              - if block.question.present?
+                %p.mb-0= block.question.content
+              - else
+                %p.mb-0.text-muted.font-italic= t('.deleted_question')
```
pixeldesu: Also use double quotes in HAML please! (see the other `t()` calls in the same template)

pixeldesu (changes requested):
I already run into errors just trying to run the migration

```
An error has occurred, this and all later migrations canceled:
PG::NotNullViolation: ERROR:  column "user_id" contains null values

migration | #<struct  ActiveRecord::MigrationProxy name="SetOnDeleteForMuteRules",  version=20220714212414,  filename="db/migrate/20220714212414_set_on_delete_for_mute_rules.rb",  scope="">
```

pixeldesu (approved):
👍 Works fine now, didn't run into any issues deleting my account or possibly seeing other account data

PR #532, opened 2022-07-18
Split up `StaticController`

raccube:
Having a "catch-all" static or pages controller is not really a good code pattern. (Despite every Rails tutorial teaching you to do it this way)


app/controllers/about_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+                        .count
+
+    @users = user_count - current_ban_count
+    @questions = Question.count
+    @answers = Answer.count
+    @comments = Comment.count
+    @smiles = Appendable::Reaction.count
+  end
+
+  def faq; end
```
pixeldesu: The FAQ page does not exist anymore, so this can be removed.

pixeldesu (changes requested):

pixeldesu (approved):

PR #538, opened 2022-07-19
Add Bootsnap

raccube:


pixeldesu:
Didn't test it yet, hope you are aware of these things for the deployment/production setup:

> Note that bootsnap writes to `tmp/cache` (or the path specified by `ENV['BOOTSNAP_CACHE_DIR']`), and that directory must be writable. Rails will fail to boot if it is not. If this is unacceptable (e.g. you are running in a read-only container and unwilling to mount in a writable tmpdir), you should remove this line or wrap it in a conditional.
>
> **Note also that bootsnap will never clean up its own cache: this is left up to you. Depending on your deployment strategy, you may need to periodically purge `tmp/cache/bootsnap*`. If you notice deploys getting progressively slower, this is almost certainly the cause.**

pixeldesu (approved):
Before (locally, WSL2):

```
processing time of bundler require: 5.193s
```

After:

```
processing time of bundler require: 1.783s
```

:shipit:

nilsding (approved):

PR #546, opened 2022-07-20
Place notifications in individual cards

raccube:


nilsding (changes requested):
missed opportunity to call the branch `cards-against-humanity`

PR #548, opened 2022-07-20
Place notifications in individual cards

raccube:
image

### DIFF
--- a/app/views/notifications/index.haml
+++ b/app/views/notifications/index.haml
@@ -1,13 +1,14 @@
 %ul.list-group#notifications
   - if @notifications.count.zero?
-    %li.list-group-item.text-center
-      .notifications--none
-        %i.fa.fa-bell-o.notification__bell-icon
-      %p
-        - if params[:type] == 'all'
-          = t(".none_filtered")
-        - else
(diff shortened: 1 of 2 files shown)


nilsding (approved):
now it's Art™

pixeldesu (changes requested):
Can't believe I have to say this, but you actually wrote illegal HTML.

> **Content model:**
> Zero or more [li](https://html.spec.whatwg.org/#the-li-element) and [script-supporting](https://html.spec.whatwg.org/#script-supporting-elements-2) elements.
> \- https://html.spec.whatwg.org/#the-ul-element

app/views/notifications/index.haml
```diff
@@ -1,17 +1,18 @@
-.card
-  %ul.list-group#notifications
-    - if @notifications.count.zero?
-      %li.list-group-item.text-center
+%ul.list-group#notifications
```
pixeldesu: If there are no `list-group-items` inside, you don't need an outside `list-group` either.

pixeldesu (changes requested):

pixeldesu (approved):

PR #550, opened 2022-07-21
Make notifications a polymorphic relation

### LINKED ISSUE #547: Decouple notification types from their target model
Currently we determine notification type based from its target model. Instead we should turn the notifications relation into a polymorphic one. This will make it possible to filter notifications by type for smiles and comment smiles again.

raccube:
Closes #547

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -1,5 +1,7 @@
+# frozen_string_literal: true
+
 class Notification < ApplicationRecord
-  belongs_to :recipient, class_name: 'User'
+  belongs_to :recipient, class_name: "User"
   belongs_to :target, polymorphic: true
 
   class << self
@@ -8,12 +10,12 @@ class Notification < ApplicationRecord
     define_cursor_paginator :cursored_for, :for
     define_cursor_paginator :cursored_for_type, :for_type
 
-    def for(recipient, options={})
-      self.where(options.merge!(recipient: recipient)).order(:created_at).reverse_order
+    def for(recipient, options = {})
+      where(options.merge!(recipient: recipient)).order(:created_at).reverse_order
     end
 
-    def for_type(recipient, type, options={})
-      self.where(options.merge!(recipient: recipient)).where(type: type).order(:created_at).reverse_order
+    def for_type(recipient, type, options = {})
+      where(options.merge!(recipient: recipient)).where(type: type).order(:created_at).reverse_order
     end
 
     def notify(recipient, target)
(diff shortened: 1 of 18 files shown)


db/migrate/20220720190421_add_type_to_notifications.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class AddTypeToNotifications < ActiveRecord::Migration[6.1]
+  def up
+    add_column :notifications, :type, :string
```
nilsding, line 5: Should probably be not nullable and have an index if you're doing queries based on this column.

db/migrate/20220720190421_add_type_to_notifications.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    execute "UPDATE notifications SET type = 'Notification::Commented' WHERE target_type = 'Comment'"
+    execute "UPDATE notifications SET type = 'Notification::QuestionAnswered' WHERE target_type = 'Answer'"
+    execute "UPDATE notifications SET type = 'Notification::StartedFollowing' WHERE target_type = 'Relationship'"
+    execute <<~SQUIRREL
```
nilsding, line 11: Oachkatzl!

app/models/notification.rb
```diff
@@ -8,12 +10,12 @@ class << self
...
 
-    def for(recipient, options={})
-      self.where(options.merge!(recipient: recipient)).order(:created_at).reverse_order
+    def for(recipient, options = {})
+      where(options.merge!(recipient: recipient)).order(:created_at).reverse_order
     end
 
-    def for_type(recipient, type, options={})
-      self.where(options.merge!(recipient: recipient)).where('LOWER(target_type) = ?', type).order(:created_at).reverse_order
+    def for_type(recipient, type, options = {})
```
nilsding: the modern Ruby thing to do would be:
```suggestion
    def for_type(recipient, type, **kwargs)
```

`kwargs` will be your usual Hash

nilsding (changes requested):

db/migrate/20220720190421_add_type_to_notifications.rb
raccube: 9103340c

app/models/notification.rb
raccube: b1380922

db/migrate/20220720190421_add_type_to_notifications.rb
raccube: :shipit:

nilsding (approved):

PR #560, opened 2022-07-23
Update `rubocop`

raccube:


pixeldesu:
This means Hound might potentially not end up working, Hound afaik only supports the version I specified at first.

Maybe it's time to switch to Reviewdog instead.

raccube:
Does reviewdog let you use a different rubocop config? We could switch to `rt_rubocop_defaults`

pixeldesu:
Reviewdogs supports whatever, it's not limited to things.

nilsding (approved):

PR #567, opened 2022-07-24
Improvements and bugfixes for questions

pixeldesu:
That PR title feels like a mobile app changelog.

**Changes:**
* **Bug fix:** Sets `direct` on direct questions again
  * **Side effect:** New direct questions don't show `1 answer` anymore
* Questions using the shared question partial actually get rendered using Markdown now
* Questions get Rails Admin links
* `direct` questions are not listed on user profiles anymore (on the `/questions` page)

Extended the UseCase tests to check for `direct` being set now.

### DIFF
--- a/app/models/user/question_methods.rb
+++ b/app/models/user/question_methods.rb
@@ -5,9 +5,9 @@ module User::QuestionMethods
 
   define_cursor_paginator :cursored_questions, :ordered_questions
 
-  def ordered_questions(author_is_anonymous: nil)
+  def ordered_questions(author_is_anonymous: nil, direct: nil)
     questions
-      .where({ author_is_anonymous: author_is_anonymous }.compact)
+      .where({ author_is_anonymous: author_is_anonymous, direct: direct }.compact)
       .order(:created_at)
       .reverse_order
   end
--- a/app/models/user/question_methods.rb
+++ b/app/models/user/question_methods.rb
@@ -7,7 +7,7 @@ module User::QuestionMethods
 
   def ordered_questions(author_is_anonymous: nil, direct: nil)
     questions
-      .where({ author_is_anonymous: author_is_anonymous, direct: direct }.compact)
+      .where({ author_is_anonymous:, direct: }.compact)
       .order(:created_at)
       .reverse_order
   end
--- a/lib/use_case/question/create.rb
+++ b/lib/use_case/question/create.rb
@@ -20,7 +20,8 @@ module UseCase
           content:             content,
           author_is_anonymous: anonymous,
           author_identifier:   author_identifier,
-          user:                source_user_id.nil? ? nil : source_user
+          user:                source_user_id.nil? ? nil : source_user,
+          direct:              true
         )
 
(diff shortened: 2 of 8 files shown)


lib/use_case/question/create.rb
```diff
@@ -20,7 +20,8 @@ def call
           content:             content,
           author_is_anonymous: anonymous,
           author_identifier:   author_identifier,
-          user:                source_user_id.nil? ? nil : source_user
+          user:                source_user_id.nil? ? nil : source_user,
+          direct:              true
```
nilsding: would be cool to have this explicitly in `CreateFollowers` as well

pixeldesu: so, that it explicitly sets `direct: false`? Because follower questions are never direct.

nilsding: exactly

app/models/user/question_methods.rb
```diff
@@ -5,9 +5,9 @@ module User::QuestionMethods
 
   define_cursor_paginator :cursored_questions, :ordered_questions
 
-  def ordered_questions(author_is_anonymous: nil)
+  def ordered_questions(author_is_anonymous: nil, direct: nil)
     questions
-      .where({ author_is_anonymous: author_is_anonymous }.compact)
+      .where({ author_is_anonymous: author_is_anonymous, direct: direct }.compact)
```
nilsding: RubyRubyRuby
```suggestion
      .where({ author_is_anonymous:, direct: }.compact)
```

lib/use_case/question/create.rb
```diff
@@ -55,7 +57,7 @@ def check_blocks
       end
 
       def increment_asked_count
-        unless source_user_id && !anonymous
+        unless source_user_id && !anonymous && !direct
           # Only increment the asked count of the source user if the question
           # is not anonymous, and we actually have a source user
```
nilsding:
```suggestion
          # is not anonymous, and is not direct, and we actually have a source user
```

lib/use_case/question/create.rb
```diff
@@ -11,6 +11,7 @@ class Create < UseCase::Base
       option :content, type: Types::Coercible::String
       option :anonymous, type: Types::Params::Bool, default: proc { false }
       option :author_identifier, type: Types::Coercible::String | Types::Nil
+      option :direct, type: Types::Params::Bool, default: proc { true }
```
nilsding, line 14: on second thought... I think this option is not needed here, as 1-1 questions (which this usecase is for) can only be direct anyways.

pixeldesu: I added the option just in case, just so we have the option to still do non-direct ones in the future, for whatever reason.

Might not be specifically needed right now, but I found this better than removing entire methods tied to other logic here.

nilsding: fair enough!

nilsding (approved):
@raccube please take a look

raccube (approved):

PR #584, opened 2022-07-27
rails_admin: group models together and add nice icons

nilsding:
Screenshot 2022-07-27 at 21 09 59

this PR basically changes the rails admin initialiser so that...

- models now have an icon next to them
- some models are now have the User or List model as parent where it makes sense
- models containing (public-facing) user-created content are separate from reports and bans

grouping and iconing TBD, but this ⬅️ is what it looks like :-)

### DIFF
--- a/config/initializers/rails_admin.rb
+++ b/config/initializers/rails_admin.rb
@@ -51,4 +51,73 @@ RailsAdmin.config do |config|
     User
     UserBan
   ]
+
+  # set up icons for some models
+  {
+    "AnonymousBlock"        => "user-secret",
+    "Answer"                => "exclamation",
+    "Appendable"            => "paperclip",
+    "Appendable::Reaction"  => "smile",
+    "Comment"               => "comment",
+    "Inbox"                 => "inbox",
+    "List"                  => "list",
+    "ListMember"            => "users",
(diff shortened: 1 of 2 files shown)


config/initializers/rails_admin.rb
```diff
@@ -51,4 +51,73 @@
...
+    "MuteRule"              => "volume-mute",
+    "Notification"          => "bell",
+    "Profile"               => "id-card",
+    "Question"              => "question",
+    "Relationship"          => "people-arrows",
+    "Relationships::Block"  => "user-slash",
+    "Relationships::Follow" => "user-friends",
+    "Report"                => "exclamation-triangle",
+    "Service"               => "network-wired",
+    "Services::Twitter"     => "dumpster-fire",
```
raccube, line 74: Would have been nice to have a Twitter icon here as it's more scanable but it seems Rails Admin doesn't include the brand icons

raccube (approved):

config/initializers/rails_admin.rb
nilsding: twitter's a dumpster fire anyway

pixeldesu (approved):

PR #592, opened 2022-07-30
Use notification types instead of target types to clear them

pixeldesu:
`target_type` is just `Appendable` so we'll have fun once comments turn into appendables as well, so we should use `type` instead.


nilsding (approved):
Die Raubkatze ist super.

raccube (approved):

PR #621, opened 2022-08-12
Site-wide anonymous blocking

### LINKED ISSUE #620: Add site-wide anonymous blocks
If _someone_ decides to do spamming in a bunch of inboxes it'd be cool to just have a global block option for moderators.

raccube:
Closes #620

### DIFF
--- a/app/controllers/ajax/anonymous_block_controller.rb
+++ b/app/controllers/ajax/anonymous_block_controller.rb
@@ -6,7 +6,7 @@ class Ajax::AnonymousBlockController < AjaxController
 
     question = Question.find(params[:question])
 
-    raise Errors::Forbidden if params[:global] && !current_user.has_role?(:moderator)
+    raise Errors::Forbidden if params[:global] && !current_user.mod?
 
     AnonymousBlock.create!(
       user:       params[:global] ? nil : current_user,
--- a/app/controllers/ajax/anonymous_block_controller.rb
+++ b/app/controllers/ajax/anonymous_block_controller.rb
@@ -11,7 +11,7 @@ class Ajax::AnonymousBlockController < AjaxController
     AnonymousBlock.create!(
       user:       params[:global] ? nil : current_user,
       identifier: question.author_identifier,
-      question:,
+      question:
     )
 
     question.inboxes.first.destroy
--- a/app/controllers/ajax/anonymous_block_controller.rb
+++ b/app/controllers/ajax/anonymous_block_controller.rb
@@ -6,10 +6,12 @@ class Ajax::AnonymousBlockController < AjaxController
 
     question = Question.find(params[:question])
 
+    raise Errors::Forbidden if params[:global] && !current_user.has_role?(:moderator)
+
     AnonymousBlock.create!(
-      user:       current_user,
+      user:       params[:global] ? nil : current_user,
       identifier: question.author_identifier,
-      question:   question
+      question:,
     )
 
     question.inboxes.first.destroy
--- a/app/controllers/ajax/anonymous_block_controller.rb
+++ b/app/controllers/ajax/anonymous_block_controller.rb
@@ -14,7 +14,7 @@ class Ajax::AnonymousBlockController < AjaxController
       question:
     )
 
-    question.inboxes.first.destroy
+    question.inboxes.first&.destroy
 
     @response[:status] = :okay
     @response[:message] = t(".success")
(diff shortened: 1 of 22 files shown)


nilsding:
PR #621 and it has a fox.  Nice.

app/controllers/ajax/anonymous_block_controller.rb
```diff
@@ -6,10 +6,12 @@ def create
 
     question = Question.find(params[:question])
 
+    raise Errors::Forbidden if params[:global] && !current_user.has_role?(:moderator)
```
nilsding: use the [`User#mod?`](https://github.com/Retrospring/retrospring/blob/41765159a9834ce184050949dd90730bee030db9/app/models/user.rb#L140) method instead of checking for the moderator role on its own

app/views/answerbox/_header.haml
```diff
@@ -18,6 +18,10 @@
                 %a.dropdown-item{ href: "#", tabindex: -1, data: { action: "ab-question-report", q_id: a.question.id } }
                   %i.fa.fa-exclamation-triangle
                   = t("voc.report")
+              - if current_user.has_role?(:moderator) && a.question.anonymous? && !a.question.generated?
```
nilsding: use the [`User#mod?`](https://github.com/Retrospring/retrospring/blob/41765159a9834ce184050949dd90730bee030db9/app/models/user.rb#L140) method instead of checking for the moderator role on its own

config/locales/frontend.en.yml
```diff
@@ -45,6 +45,10 @@ en:
       confirm:
         title: "Are you sure?"
         text: "This question will be gone forever."
+    mod_mute:
+      confirm:
+        title: "Are you sure?"
+        text: "This will block this user for everyone."
```
nilsding: would say something like:
```suggestion
        text: "This will mute this anonymous coward for everyone."
```

nilsding (changes requested):

app/views/answerbox/_header.haml
```diff
@@ -18,6 +18,10 @@
...
+                %a.dropdown-item{ name: "mod-block-anon", data: { q_id: a.question.id } }
```
pixeldesu:
```suggestion
                %a.dropdown-item{ href: "#", name: "mod-block-anon", data: { q_id: a.question.id } }
```

Links without `href` are not valid.

app/views/inbox/_entry.haml
```diff
@@ -22,15 +22,18 @@
             .dropdown-menu.dropdown-menu-right{ role: :menu }
               - if i.question.user_id != current_user.id
                 %a.dropdown-item{ name: "ib-report", data: { q_id: i.question.id } }
-                  %i.fa.fa-warning
+                  %i.fa.fa-fw.fa-warning
                   = t("voc.report")
-              - if i.question.author_is_anonymous && i.question.author_identifier.present?
+              - if current_user.has_role?(:moderator) && i.question.anonymous? && !i.question.generated?
```
pixeldesu: use the [`User#mod?`](https://github.com/Retrospring/retrospring/blob/41765159a9834ce184050949dd90730bee030db9/app/models/user.rb#L140) method instead of checking for the moderator role on its own

pixeldesu (changes requested):
Looking at the code this seems fine generally, but with the currently wrong permission check it's hardly testable as an admin account.

config/locales/frontend.en.yml
pixeldesu: I agree with the mute suggestion here, the anonymous coward part not so much :P

pixeldesu (changes requested):
Things are visible in the UI now, but I get this error when it tries to create an `AnonymousBlock`:

```
PG::NotNullViolation: ERROR:  null value in column "user_id" violates not-null constraint
DETAIL:  Failing row contains (139, 93f499adb76c80f8a57abdee9eab62b55de560bccba96b82cc639f911530287f..., 2022-08-17 21:39:38.049285, 2022-08-17 21:39:38.049285, null, 108830372973194772).

```

app/javascript/retrospring/features/moderation/blockAnon.ts
```diff
@@ -0,0 +1,44 @@
+import Rails from '@rails/ujs';
+import swal from 'sweetalert';
+
+import { showErrorNotification, showNotification } from "utilities/notifications";
+import I18n from "retrospring/i18n";
+
+export function blockAnonEventHandler(event: Event): void {
```
pixeldesu, line 7:
```suggestion
export function blockAnonEventHandler(event: Event): void {
    event.preventDefault();
    
```

now that the link is a proper `a` tag we need this.

pixeldesu (changes requested):

pixeldesu:
Another thing, this also might cause issues if we block someone from the feed instead of the inbox.
```
NoMethodError at /ajax/block_anon
=================================

undefined method `destroy' for nil:NilClass
    question.inboxes.first.destroy
```

pixeldesu (approved):

PR #622, opened 2022-08-14
typoed_email_validator: add new endings to the typo list

nilsding:


app/validators/typoed_email_validator.rb
```diff
@@ -20,13 +20,16 @@ class TypoedEmailValidator < ActiveModel::EachValidator
...
       gmaill.com
       gmali.com
+      gnail.com
       hotamil.com
       hotmai.com
       hotmaill.com
       iclooud.com
       iclould.com
       icluod.com
+      xn--gmail-xk1c.com
```
nilsding, line 30: I have no idea how someone managed to sign up with that.  It could be that browser seems to do some substitution in the domain...

Firefox 103 doesn't do the substitution, Safari 15.6 doesn't allow emoji in email (it asks you to "Enter an email address"), and other browsers do not exist on my machines to test.  So what the hell.

pixeldesu (approved):

PR #630, opened 2022-08-17
List direct questions for current users or mods

pixeldesu:
Instead of completely removing direct questions from user pages, this now allows moderators (in unmask mode) and users that own the current page to view their direct questions.

### DIFF
--- a/app/helpers/user_helper.rb
+++ b/app/helpers/user_helper.rb
@@ -21,6 +21,10 @@ module UserHelper
     current_user&.mod? && session[:moderation_view] == true
   end
 
+  def belongs_to_current_user?
+    @user == current_user
+  end
+
   private
 
   def profile_link(user)
--- a/app/helpers/user_helper.rb
+++ b/app/helpers/user_helper.rb
@@ -21,10 +21,6 @@ module UserHelper
     current_user&.mod? && session[:moderation_view] == true
   end
 
-  def belongs_to_current_user?
-    @user == current_user
-  end
-
   private
 
   def profile_link(user)
--- a/app/helpers/user_helper.rb
+++ b/app/helpers/user_helper.rb
@@ -18,7 +18,7 @@ module UserHelper
   end
 
   def moderation_view?
-    current_user&.mod? && session[:moderation_view] == true
+    !!(current_user&.mod? && session[:moderation_view] == true)
   end
 
   private
--- a/spec/controllers/user_controller_spec.rb
+++ b/spec/controllers/user_controller_spec.rb
@@ -50,15 +50,57 @@ describe UserController, type: :controller do
   end
 
   describe "#questions" do
+    let!(:question) { FactoryBot.create(:question, user: user, direct: true, author_is_anonymous: false) }
     subject { get :questions, params: { username: user.screen_name } }
 
-    context "user signed in" do
-      before(:each) { sign_in user }
+    it "renders the user/questions template" do
+      subject
+      expect(assigns(:user)).to eq(user)
+      expect(response).to render_template("user/questions")
+    end
 
-      it "renders the user/questions template" do
+    context "current user signed in" do
+      before do
+        sign_in user
+      end
+
+      it "renders all questions" do
         subject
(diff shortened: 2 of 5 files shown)


app/helpers/user_helper.rb
```diff
@@ -21,6 +21,10 @@ def moderation_view?
     current_user&.mod? && session[:moderation_view] == true
   end
 
+  def belongs_to_current_user?
```
nilsding: This method would be better suited as a private method inside UserController; helpers should be for views only.

btw, you can also use Ruby3's endless method definition syntax:
```ruby
private

def belongs_to_current_user? = @user == current_user
```

nilsding (changes requested):

nilsding (approved):

pixeldesu:
So I adjusted this, but I didn't manage to properly get tests done for `UserController#questions`. Here's my attempt in a gist: https://gist.github.com/pixeldesu/f80868e0c196227c0adb3e6780ae31b9

The first test was similar to the others but that didn't work either and checking for the size of the instance variable doesn't yield the right answer too, even though it works fine as expected in the frontend. (the checks for no renders/null length pass because the first one fails, etc. etc.)

nilsding:
it's because you used `let` instead of `let!`; `let!` expressions are evaluated eagerly whereas `let` ones will get evaluated the first time they're called

pixeldesu:
It wasn't working with a question being created in a `before` block either. And I just tried it, `@question.size` still is 0 with `let!`.

nilsding:
Ah, found it.  The question is created as an anonymous one.  Pass `author_is_anonymous: false` to the `FactoryBot.create` call and then it does what it should.

pixeldesu:
Yup, that fixed that.

I will adjust the other tests (the moderation one wasn't properly implemented yet) and then push those. Wanted to have this behaviour tested properly because it could lead to unwanted infos being leaked.

spec/controllers/user_controller_spec.rb
```diff
@@ -50,15 +52,57 @@
...
+    let!(:question) { FactoryBot.create(:question, user:, direct: true, author_is_anonymous: false) }
     subject { get :questions, params: { username: user.screen_name } }
 
+    it "renders the user/questions template" do
+      subject
+      expect(assigns(:user)).to eq(user)
+      expect(response).to render_template("user/questions")
+    end
+
+    context "current user signed in" do
```
nilsding:
```suggestion
    context "when the current user is the question author" do
```

spec/controllers/user_controller_spec.rb
```diff
@@ -50,15 +52,57 @@
...
+        sign_in user
+      end
+
+      it "renders all questions" do
+        subject
+        expect(assigns(:questions).size).to eq(1)
+      end
+    end
+
     context "user signed in" do
```
nilsding:
```suggestion
    context "when the current user is someone else" do
```

spec/controllers/user_controller_spec.rb
```diff
@@ -50,15 +52,57 @@
...
-      it "renders the user/questions template" do
+      it "renders no questions" do
         subject
-        expect(assigns(:user)).to eq(user)
-        expect(response).to render_template("user/questions")
+        expect(assigns(:questions).size).to eq(0)
+      end
+    end
+
+    context "moderator unmasked" do
```
nilsding:
```suggestion
    context "when a moderator uses unmask" do
```

spec/controllers/user_controller_spec.rb
```diff
@@ -50,15 +52,57 @@
...
+        allow_any_instance_of(UserHelper).to receive(:moderation_view?) { true }
+      end
+
+      it "contains all questions" do
+        subject
+        expect(assigns(:questions).size).to eq(1)
+      end
+    end
+
+    context "user not signed in" do
```
nilsding:
```suggestion
    context "when user is not signed in" do
```

nilsding (approved):

nilsding (approved):

PR #631, opened 2022-08-19
answerbox: use ActiveSupport's String#truncate

nilsding:
[ActiveSupport's `#truncate` does a bit more than just cutting off a string](https://github.com/rails/rails/blob/2f9c84a604b3288504196e23c95348221a298b35/activesupport/lib/active_support/core_ext/string/filters.rb#L48).  there's also a [view helper for truncate](https://github.com/rails/rails/blob/2f9c84a604b3288504196e23c95348221a298b35/actionview/lib/action_view/helpers/text_helper.rb#L98), but I couldn't get that to work nicely with the "read more" button and the markdown thing.  

## before

Screenshot 2022-08-19 at 16 36 43

## after

Screenshot 2022-08-19 at 16 36 16

### DIFF
--- a/app/views/application/_answerbox.haml
+++ b/app/views/application/_answerbox.haml
@@ -5,9 +5,8 @@
   .card-body
     - if display_all.nil?
       .answerbox__answer-text
-        = markdown a.content[0..640]
+        = markdown a.content.truncate(640, omission: " [...]", separator: /\s/)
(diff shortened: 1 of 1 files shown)


nilsding:
~~ah, actually I use the view helper here, I meant passing an optional block to it.~~

pixeldesu (approved):

PR #633, opened 2022-08-20
Move actions from `ModerationController` to `Moderation::ReportsController`

pixeldesu:
More preparation to actually get a proper moderation panel done at some time, just extracting report-view related actions into `Moderation::ReportsController`.


spec/controllers/moderation/reports_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  let(:user) { FactoryBot.create :user, roles: ["moderator"] }
+
+  describe "#index" do
+    subject { get :index }
+
+    before do
+      sign_in user
+    end
+
+    it "renders the user/questions template" do
```
nilsding:
```suggestion
    it "renders the moderation/reports/index template" do
```

nilsding (approved):

raccube (approved):

PR #634, opened 2022-08-21
Add view to show all questions from one author by identifier

### LINKED ISSUE #582: Add moderation view to display all questions from an anonymous user
I'd like to be able to quickly view all questions sent by someone with an `author_identifier`, accessible via the dropdown on a question.

Possible URL could be `/moderation/outbox/{identifier}`

pixeldesu:
Since not everyone is an actual user, having a proper view for this is quite nice. This is accessible via the actions dropdown on anonymous questions.

Local example:
image

Fixes #582


app/controllers/moderation/questions_controller.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Moderation::QuestionsController < ApplicationController
+  before_action :authenticate_user!
+
+  def index
```
nilsding: would be better to call this method `show` as it needs an id (`index` would be a listing of all the things instead of a specific one)

config/routes.rb
```diff
@@ -26,6 +26,7 @@
     get "/moderation/blocks", to: "moderation/anonymous_block#index", as: :mod_anon_block_index
     get "/moderation/reports(/:type)", to: "moderation/reports#index", as: :moderation_reports, defaults: { type: "all" }
     get "/moderation/inbox/:user", to: "moderation/inbox#index", as: :mod_inbox_index
+    get "/moderation/questions/:author_identifier", to: "moderation/questions#index", as: :moderation_questions
```
nilsding:
```suggestion
    get "/moderation/questions/:author_identifier", to: "moderation/questions#show", as: :moderation_questions
```

raccube (approved):

app/views/moderation/questions/_header.haml
```diff
@@ -0,0 +1,6 @@
+.card.question--fixed{ class: hidden ? 'question--hidden' : '', tabindex: hidden ? -1 : '', aria: { hidden: hidden } }
+  .container
+    .card-body
+      .media
+        .media-body
+          = t(".title_html", author_identifier: author_identifier, short: author_identifier.truncate(32))
```
nilsding: nieuwe regel ontbreekt

nilsding (changes requested):

nilsding (approved):

PR #635, opened 2022-08-21
Split actions into separate partials

pixeldesu:
Especially with questions and the several places they are shown in the actions that can be taken on them have been all over the place.

This PR splits the actions for answers/questions/comments into a `actions/` folder, so the dropdowns can simply be referenced as partials and edited in one place.

Fixed icon alignment as well as the fact that we accidentally made anon-blocking mod only (not the global one)


raccube:
image

(Yes, I re-exported)

raccube (approved):

PR #658, opened 2022-09-03
Refactor inbox TS functionality to use `@rails/request.js`

pixeldesu:
More request.js refactoring

**Testing:**

* Test all inbox functionality (answering, deleting questions, anon blocking, generating etc.)


raccube (approved):
Looks fine except one minor thing: The 'Load more' button stays when you delete all inbox entries.

pixeldesu:
I think messing with the load more button here would break the page if the user decided to stay on it. I'd wait out on this until we refactor pagination in general.

PR #660, opened 2022-09-03
Refactor moderation TS functionality to use `@rails/request.js`

pixeldesu:
More request.js refactoring

**Testing:**
* Test moderation functionality (global anon blocks, removing reports, banning, user privilege...)


raccube (approved):
Missing translation when banning but otherwise ok

PR #663, opened 2022-09-04
Replace Turbolinks with Turbo

### LINKED ISSUE #412: Switch from Turbolinks to Turbo
[Turbolinks has been discontinued](https://github.com/turbolinks/turbolinks) and replaced by [Turbo Drive](https://turbo.hotwired.dev/handbook/drive).

This probably has less issues with current Rails versions and tooling compared to Turbolinks, so it's worth to switch to Turbo instead. Especially because of the other parts like Turbo Frames and Turbo Streams as well, improving interactivity with us having to write less JS at all.

pixeldesu:
Fixes #412

Turbo changes up things quite massively compared to Turbolinks. All form submissions are now caught by it, requiring us to `data-turbo="false"` every form that does not support Turbo-behavior at all, which is basically all of settings.

**Testing:**
* Everything, please. Just make sure things work fine.


raccube:
Removing 2FA doesn't trigger a confirmation dialog but fires off the request without reloading the page.

raccube (approved):
Testing notes:

```
unbanning
following
unfollowing
asking
answering
reacting a
commenting
reacting c
q gen
clearing inbox
notifications ok
profiles look ok
updating account
updating profile
changing avatar
setting and deleting theme
unblocking anon
blocking anon
blocking user
unblocking user
removing 2fa
adding 2fa
pagination
mute rules add/remove
exporting account

PR #678, opened 2022-09-11
WebPush support

raccube:


pixeldesu:
Migrating doesn't work.

```
app@0e014e1fb0fb:/app$ bundle exec rails db:migrate
processing time of bundler require: 1.861s
processing time of routes.rb: 0.114s
processing time of routes.rb: 0.041s
processing time of Rails.application.initialize!: 0.743s
== 20220909161542 AddRpush: migrating =========================================
rails aborted!
StandardError: An error has occurred, this and all later migrations canceled:

uninitialized constant AddRpush::CreateRapnsNotifications

    [CreateRapnsNotifications, CreateRapnsFeedback,
     ^^^^^^^^^^^^^^^^^^^^^^^^
/app/db/migrate/20220909161542_add_rpush.rb:26:in `migrations'
/app/db/migrate/20220909161542_add_rpush.rb:33:in `up'

Caused by:
NameError: uninitialized constant AddRpush::CreateRapnsNotifications

    [CreateRapnsNotifications, CreateRapnsFeedback,
     ^^^^^^^^^^^^^^^^^^^^^^^^
/app/db/migrate/20220909161542_add_rpush.rb:26:in `migrations'
/app/db/migrate/20220909161542_add_rpush.rb:33:in `up'
Tasks: TOP => db:migrate
(See full trace by running task with --trace)
```

db/migrate/20220909161542_add_rpush.rb
```diff
@@ -0,0 +1,338 @@
...
+  end
+
+  def self.down
+    migrations.reverse.each do |m|
+      m.down
+    rescue ActiveRecord::StatementInvalid => e
+      Rails.logger.debug e
+    end
+  end
+
```
nilsding, line 44:
```suggestion

  # rubocop:disable Rails/MigrationClassName

```

db/migrate/20220909161542_add_rpush.rb
```diff
@@ -0,0 +1,338 @@
...
+  class AddFailAfterToRpushNotifications < ActiveRecord::Migration[5.0]
+    def self.up
+      add_column :rpush_notifications, :fail_after, :timestamp, null: true
+    end
+
+    def self.down
+      remove_column :rpush_notifications, :fail_after
+    end
+  end
```
nilsding, line 338:
```suggestion
  end

  # rubocop:enable all
```

pixeldesu (changes requested):
Enabling push notifications and receiving them works fine.

Tested on Windows 11, with:
* Firefox 108.0.1
* Microsoft Edge 108.0.1462.54

----

Push notifications from anonymous accounts just have "New question from" as a title, which is kind of broken.

Other than that, it looks fine to me. Later we maybe should show the avatar of users in question notifications, but for now the notification looks good like this.

----

Also please fix the remaining code style lints in some of the files (excluding ABC-size, method length, etc. there are some layout/indent issues)

app/workers/push_notification_worker.rb
```diff
@@ -0,0 +1,17 @@
...
+  include Sidekiq::Worker
+
+  sidekiq_options queue: :push_notification, retry: 0
+
+  def perform(notification_id)
+    Rpush.config.push = true
+    Rpush::Daemon.common_init
+    Rpush::Daemon::Synchronizer.sync
+    Rpush::Daemon::AppRunner.enqueue(Rpush::Client::ActiveRecord::Notification.where(id: notification_id))
+    Rpush::Daemon::AppRunner.stop
```
nilsding: hmm...  I wonder if it would be better to put some of these in the sidekiq initializer, e.g.
```ruby
Sidekiq.configure_server do
    Rpush.config.push = true
    Rpush::Daemon.common_init
    Rpush::Daemon::Synchronizer.sync
end

at_exit do
    Rpush::Daemon::AppRunner.stop
end
```

and just leave the enqueue bit in the job.

sounds like these start/stop global instances of some kind which might cause some issues if two of these jobs run at the same time in the same sidekiq process

app/controllers/ajax/web_push_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def unsubscribe
+    params.permit(:endpoint)
+
+    removed = if params.key?(:endpoint)
+      current_user.web_push_subscriptions.where("subscription ->> 'endpoint' = ?", params[:endpoint]).destroy_all
+    else
+      current_user.web_push_subscriptions.destroy_all
+    end
```
nilsding: to make rubocop happy:

```suggestion
    removed = if params.key?(:endpoint)
                current_user.web_push_subscriptions.where("subscription ->> 'endpoint' = ?", params[:endpoint]).destroy_all
              else
                current_user.web_push_subscriptions.destroy_all
              end
```

you can also yeet this `if`-block into a private function and call it instead, it might make the AbcSize lower (but it's fine by me this way)

app/controllers/ajax/web_push_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      user:         current_user,
+      subscription: params[:subscription]
+    )
+
+    @response[:status] = :okay
+    @response[:success] = true
+    @response[:message] = t(".subscription_count", count: current_user.web_push_subscriptions.count)
+  end
+
+  def unsubscribe
```
nilsding:
```suggestion
  def unsubscribe # rubocop:disable Metrics/AbcSize
```

db/migrate/20220909220449_add_webpush_app.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require "webpush"
+
+class AddWebpushApp < ActiveRecord::Migration[6.1]
+  def up
+    vapid_keypair = Webpush.generate_key.to_hash
+    app = Rpush::Webpush::App.new
+    app.name = "webpush"
+    app.certificate = vapid_keypair.merge(subject: "user@example.com").to_json # TODO: put an email address here
```
nilsding: TODO????

config/environments/development.rb
```diff
@@ -71,8 +71,9 @@
   # Use an evented file watcher to asynchronously detect changes in source code,
   # routes, locales, etc. This feature depends on the listen gem.
   # config.file_watcher = ActiveSupport::EventedFileUpdateChecker
+  config.hosts << `hostname`.chomp
```
nilsding: please add a comment why this is done here

lib/use_case/question/create.rb
```diff
@@ -33,6 +33,9 @@ def call
 
         inbox = ::Inbox.create!(user: target_user, question: question, new: true)
 
+        webpush_app = ::Rpush::App.find_by(name: "webpush")
+        target_user.push_notification(webpush_app, inbox)
```
nilsding: might be better to do this instead so that it doesn't break if the rpush-app is not there yet:

```suggestion
        target_user.push_notification(webpush_app, inbox) if webpush_app
```

app/workers/question_worker.rb
```diff
@@ -10,13 +10,15 @@ class QuestionWorker
...
+    webpush_app = Rpush::App.find_by(name: "webpush")
 
     user.followers.each do |f|
       next if f.inbox_locked?
       next if f.banned?
       next if MuteRule.where(user: f).any? { |rule| rule.applies_to? question }
 
-      Inbox.create(user_id: f.id, question_id: question_id, new: true)
+      inbox = Inbox.create(user_id: f.id, question_id:, new: true)
+      f.push_notification(webpush_app, inbox)
```
nilsding: might be better to do this instead so that it doesn't break if the rpush-app is not there yet

```suggestion
      f.push_notification(webpush_app, inbox) if webpush_app
```

nilsding (changes requested):
looks promising overall!

db/migrate/20220909220449_add_webpush_app.rb
raccube: Oops. Missed that.

config/environments/development.rb
raccube: Was for local env, should not have committed this.

public/service_worker.js
```diff
@@ -0,0 +1,23 @@
+self.addEventListener('push', function (event) {
+  if (event.data) {
+    const notification = event.data.json();
+    console.log(event.data);
```
pixeldesu:
```suggestion
```

We don't `console.log` in this house, except for errors.

pixeldesu (approved):
LGTM :shipit: (please remove my nit and fix some lints though)

db/migrate/20220909220449_add_webpush_app.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require "webpush"
+
+class AddWebpushApp < ActiveRecord::Migration[6.1]
+  def up
+    vapid_keypair = Webpush.generate_key.to_hash
+    app = Rpush::Webpush::App.new
+    app.name = "webpush"
+    app.certificate = vapid_keypair.merge(subject: APP_CONFIG["contact_email"]).to_json
```
nilsding: please make it fail if it's unset

```suggestion
    app.certificate = vapid_keypair.merge(subject: APP_CONFIG.fetch("contact_email")).to_json
```

nilsding (approved):

PR #689, opened 2022-09-17
Update i18n-js to 4.x

pixeldesu:
_I love breaking changes everywhere!_

> **Warning**
> This requires a change in the deployment after being merged

This pretty much changes up how `i18n-js` works in most parts configuration-wise. A new (non-rake) CLI, new configuration (including default path) that dropped 90% of the available options, and the JS side also changing how it works.

Anyway, works now.

**Testing:**
```
bundle i18n export
```
and then try to open a confirm dialog, (un)subscribing to notifications or anything else that has frontend locales.


Gemfile.lock
```diff
@@ -326,7 +324,12 @@ GEM
     notiffany (0.1.3)
       nenv (~> 0.1)
       shellany (~> 0.0)
-    oauth (0.5.8)
+    oauth (1.1.0)
```
nilsding, line 320: where did this upgrade come from?

pixeldesu: Probably because I `bundle update`d (Rails got updated to what Dependabot suggested as well)

pixeldesu:
Superseeded by #727

PR #727, opened 2022-10-10
Install i18n-js 4 2: Electric Boogaloo

pixeldesu:
_I love breaking changes everywhere! **Now without additional dependency updates!**_

> **Warning**
> This requires a change in the deployment after being merged

This pretty much changes up how `i18n-js` works in most parts configuration-wise. A new (non-rake) CLI, new configuration (including default path) that dropped 90% of the available options, and the JS side also changing how it works.

Anyway, works now.

**Testing:**
```
bundle i18n export
```
and then try to open a confirm dialog, (un)subscribing to notifications or anything else that has frontend locales.

### DIFF
--- a/.github/workflows/retrospring.yml
+++ b/.github/workflows/retrospring.yml
@@ -61,7 +61,7 @@ jobs:
           npm i -g yarn
           yarn install --frozen-lockfile
       - name: Export i18n JS files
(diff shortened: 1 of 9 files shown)


raccube (approved):
Deployment has been adjusted

PR #728, opened 2022-10-10
Use appropriate image sizes and lazy loading

pixeldesu:
This probably will improve loading performance drastically. Uses image sizes that actually make sense and implements `loading="lazy"` where possible.

### DIFF
--- a/app/views/question/_question.haml
+++ b/app/views/question/_question.haml
@@ -4,7 +4,7 @@
       .media
         - unless question.author_is_anonymous
           %a.pull-left{ href: unless hidden then user_path(question.user) end }
-            %img.answerbox__question-user-avatar.avatar-md{ src: question.user.profile_picture.url(:medium) }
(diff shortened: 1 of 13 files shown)


app/views/question/_question.haml
```diff
@@ -4,7 +4,7 @@
       .media
         - unless question.author_is_anonymous
           %a.pull-left{ href: unless hidden then user_path(question.user) end }
-            %img.answerbox__question-user-avatar.avatar-md{ src: question.user.profile_picture.url(:medium) }
+            %img.answerbox__question-user-avatar.avatar-md{ src: question.user.profile_picture.url(:small) }
```
raccube, line 7: This one is missing a lazy load. Is this intentional?

pixeldesu: Yes, because it's always above the fold, so it doesn't require to be lazy-loaded.

raccube (approved):

PR #730, opened 2022-10-11
auto-focus text field in ask modal

### LINKED ISSUE #557: Auto foxus text field in ask modal

nilsding:
resolves #557

unfortunately this needed some jQuery call as bootstrap 4 still relies on it heavily, but it works!


raccube (approved):
Tested with Firefox/macOS. 
On Safari/iOS, it focuses the field but doesn't bring up the keyboard. Don't think there's anything we can do with that.

pixeldesu (approved):

PR #738, opened 2022-10-16
Replace inbox sidebar with action bar

pixeldesu:
The sidebar layout isn't particularly great, especially because it doesn't contain that much info or actions for the size it takes.

The action bar at the top uses way less space, and is in front of the inbox content on mobile, compared to at the bottom with the sidebar. A lot of people might have missed some inbox functionality through this.

|         | Before | After |
|---------|--------|-------|
| Mobile  | image | image |
| Desktop | image | image |


raccube (changes requested):
I like this but I don't think it's very clear.

Here's a few things I think would make it better:
- Change the placeholder text on the user filter to "Filter by user…"
- Change the "Get new question" button text to "Generate question".

app/views/inbox/_actions.haml
```diff
@@ -0,0 +1,23 @@
...
+            %i.fa.fa-fw.fa-share-alt
+            %span.sr-only= t(".share.heading")
+          .dropdown-menu.dropdown-menu-right{ role: :menu }
+            %a.dropdown-item{ href: "https://twitter.com/intent/tweet?text=Ask%20me%20anything%21&url=#{user_url(current_user)}", target: "_blank" }
+              %i.fa.fa-fw.fa-twitter
+              = t(".share.button", service: "Twitter")
+            %a.dropdown-item{ href: "https://www.tumblr.com/share/link?url=#{user_url(current_user)}&name=Ask%20me%20anything%21", target: "_blank" }
+              %i.fa.fa-fw.fa-tumblr
+              = t(".share.button", service: "Tumblr")
+        %button.btn.btn-danger.ml-1{ type: :button, id: delete_id, disabled: (disabled ? :disabled : nil), data: { ib_count: inbox_count } }
```
nilsding:
```suggestion
        %button.btn.btn-danger.ml-1{ type: :button, title: t(".actions.delete"), id: delete_id, disabled: (disabled ? :disabled : nil), data: { ib_count: inbox_count } }
```

app/views/inbox/_actions.haml
```diff
@@ -0,0 +1,23 @@
...
+  .card-body
+    .d-md-flex
+      = bootstrap_form_tag url: inbox_path, method: :get, html: { class: "d-block" } do |f|
+        = f.text_field :author, value: params[:author], placeholder: t(".author.placeholder"), prepend: "@", hide_label: true,
+          append: f.button(t(".author.button"), name: nil, class: "btn btn-light", id: "ib-author")
+      .d-flex.ml-auto.mt-2.mt-md-0
+        %button.btn.btn-info.mr-auto{ type: :button, id: "ib-generate-question" }
+          = t(".questions.button")
+        .button-group.ml-1
+          %button.btn.btn-default{ data: { toggle: :dropdown }, aria: { expanded: false } }
```
nilsding:
```suggestion
          %button.btn.btn-default{ title: t(".share.heading"), data: { toggle: :dropdown }, aria: { expanded: false } }
```

so it has a hover tooltip, just the icon is a bit too less imho

nilsding:
I like the general idea of this, but I feel like it's taking up a bit too much space.  Adding `padding: .5rem 1.25rem;` to the `.card-body` styling of the toolbar makes it a bit more comfy I think:
 
Screenshot 2022-10-22 at 01 00 10

raccube (approved):
With the exception of @nilsding's comments, I think it looks good.

pixeldesu:
@nilsding I have made a compromise, I decreased the vertical spacing, but not as much as you suggested.

image

This adjustment is `padding: .75rem 1.25rem;` so the spacing matches Bootstraps `card-header` which we also use in answerboxes and inbox entries.

nilsding:
Jo passt.

side note: I don't like how the share button looks more like just an icon than a button, but this is a general bootstrap thingy and out of scope for this PR

nilsding (approved):

pixeldesu:
Hmm, I just didn't want to give it any actionable button color that gives it some sort of urgency

I could reorder it to share, generate, delete so it doesn't look that much out of place.

nilsding:
all I'm missing is some borders :D  but for me the order is good as it is now

PR #748, opened 2022-10-20
Use data-turbo-method for mod (un)masking link

pixeldesu:
This fixes the 404 when clicking the _Moderation View_ links in the navigation dropdown.

It does not fix that now the redirect isn't working (it triggers a Turbo Streams response) but I haven't found a way to work around that yet, for now this is better than a breaking functionality.


raccube (approved):

nilsding (approved):
LGTM, please create a follow-up issue so we won't forget about the broken redirect

PR #758, opened 2022-10-24
allow for using SVG logos in the navbar

nilsding:
this uses our ✨new and improved✨ logo by default, but can be easily changed by modifying `/public/logo.svg`.  the svg is loaded only once when the application starts, so any modifications to it after a deployment are visible immediately.

thanks to the power of CSS and SVG being able to make use of it, custom colour schemes still work too.

bildnis sonst ward ihm kein glaube geschenkt (im feuerfuchs und safari):
Screenshot 2022-10-24 at 22 43 50


pixeldesu (approved):
👍

PR #775, opened 2022-10-31
Remove Guard and Brakeman

pixeldesu:
I'm pretty sure we haven't used that _ever_.

And now we have CodeQL to notify us of code issues with CVEs and such.


nilsding (approved):
Spiderman

PR #778, opened 2022-11-03
Update Font Awesome to version 6

### LINKED ISSUE #400: Upgrade FontAwesome 4 to FontAwesome 6
In order to use the new optimized assets and have a wider variety of icons, we should upgrade to the latest version of FontAwesome.

**Areas:**
* Views
* Ruby (Helpers)

pixeldesu:
This was easier than I thought, mainly coming from the fact that even version 6 of Font Awesome has shims for icons from version 4, which we can use directly.

Also had to re-add the `pull-*` classes because they apparently were part of FA 4.


raccube (approved):
Looks good, here are a few icon substitution suggestions:

- `fa-user-secret` → `fa-mask`
- `fa-volume-off` → `fa-user-slash` or `fa-user-shield`

PR #779, opened 2022-11-06
Switch to Node 14

pixeldesu:
Smaller step instead of trying to go to 16 like last time, maybe CI builds this and we can update Node on our server.


nilsding (approved):
updated to node 14 on the hosts

PR #780, opened 2022-11-06
Add inbox locking for users

pixeldesu:
This allows users to lock their inbox from receiving any questions. They will not receive questions from people they follow nor will they be able to get questions asked on their profile.

This feature can be used to stop a wave of harassment entirely or if someone received a lot of questions and wants to work through their inbox first.

Also a lot of people wanted to temporarily close their account and...in the process deleted it fully, so this is another circumvention for that.

Screenshot of profile with inbox locking:
image

**Testing:**
* Lock your inbox (in privacy settings) and see the message on your profile
* Send a question to followers, with one having their inbox locked, who then does not have that question in their inbox.

### DIFF
--- a/app/controllers/settings/privacy_controller.rb
+++ b/app/controllers/settings/privacy_controller.rb
@@ -6,7 +6,8 @@ class Settings::PrivacyController < ApplicationController
   def edit; end
 
   def update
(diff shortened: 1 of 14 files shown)


raccube (approved):

nilsding (approved):
neat!

PR #795, opened 2022-11-13
Add user requirement option for users

pixeldesu:
To prevent harassment a bit more users can now require users to have an account to ask (anonymous) questions.

image

### DIFF
--- a/lib/errors.rb
+++ b/lib/errors.rb
@@ -44,14 +44,17 @@ module Errors
     end
   end
 
+  class NotAuthorized < Base
+    def status
+      401
+    end
+  end
+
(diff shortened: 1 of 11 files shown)


lib/errors.rb
```diff
@@ -51,6 +51,9 @@ class UserNotFound < NotFound
   class Blocked < Forbidden
   end
 
+  class LoginRequired < Forbidden
```
nilsding: i guess a new base error `NotAuthorized` with a 401 status would make more sense for that one

nilsding (approved):

raccube (approved):

PR #806, opened 2022-11-15
Add setting to hide a users social graph

### LINKED ISSUE #543: Add option to hide a users social graph
_social graph = followers, followings_

Just an improvement for user privacy.

pixeldesu:
If someone doesn't want to show their followers/followings to the world!

The current user can view their own followings still, guests and other users won't see the tabs. Trying to access the views by URL redirects to the main profile page.

**Screenshot:**
image

**Testing:**
* Test the setting if the counts are hidden properly and that accessing the views does not work.

Closes #543

### DIFF
--- a/config/locales/activerecord.en.yml
+++ b/config/locales/activerecord.en.yml
@@ -72,7 +72,7 @@ en:
         privacy_require_user: "Require users to be logged in to ask you questions"
         privacy_allow_public_timeline: "Show your answers in the public timeline"
         privacy_allow_stranger_answers: "Allow other people to answer your questions"
-        privacy_hide_social_graph: "Hide your social graph"
+        privacy_hide_social_graph: "Hide your social graph from others"
         profile_picture: "Profile picture"
         profile_header: "Profile header"
         sign_in_count: "Sign in count"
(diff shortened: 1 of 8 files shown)


config/locales/activerecord.en.yml
```diff
@@ -72,6 +72,7 @@ en:
         privacy_require_user: "Require users to be logged in to ask you questions"
         privacy_allow_public_timeline: "Show your answers in the public timeline"
         privacy_allow_stranger_answers: "Allow other people to answer your questions"
+        privacy_hide_social_graph: "Hide your social graph"
```
nilsding: would be more clear with that
```suggestion
        privacy_hide_social_graph: "Hide your social graph from others"
```

without that I would expect to not even be able to see it myself

db/schema.rb
```diff
@@ -228,6 +228,7 @@
     t.integer "raised_accent", default: 16250871
     t.integer "light_color", default: 16316922
     t.integer "light_text", default: 0
+    t.integer "input_placeholder", default: 7107965, null: false
```
nilsding: 👀

nilsding (approved):
looks nice.  thanks for implementing all of these recent privacy features btw!

raccube (approved):

PR #809, opened 2022-11-17
Generate inbox entries using Turbo Streams

pixeldesu:
No need for AJAx and a bunch of TypeScript to handle this!

**Testing:**
* Try to generate an inbox entry


app/controllers/ajax/inbox_controller.rb
```diff
@@ -1,25 +1,4 @@
...
-                                author_identifier:   "justask",
-                                user:                current_user)
-
-    inbox = Inbox.create!(user: current_user, question_id: question.id, new: true)
-
-    @response[:status] = :okay
-    @response[:message] = t(".success")
-    @response[:success] = true
-    @response[:render] = render_to_string(partial: 'inbox/entry', locals: { i: inbox })
-    inbox.update(new: false)
```
nilsding, line 20: This is missing from the new implementation, otherwise the generated question stays read until the inbox page is refreshed again.

Maybe we can create it with `new: false` and have some "magic" variable that renders it to look like it's new

nilsding (changes requested):

app/controllers/ajax/inbox_controller.rb
pixeldesu: Hmm, setting it in the partial would work, but that is kinda ugly.

pixeldesu: Fixed!

nilsding: works as it used to, nice!

nilsding (approved):

raccube (approved):

PR #811, opened 2022-11-18
Migrate mute rule handling to Turbo Streams

### LINKED ISSUE #80: Implement Policies
Currently, a lot of the code that authorizes or prevents actions from happening or being displayed is based on long chains of conditions due to the nature of privacy settings, anonymity, etc.

Instead of managing the authorization inside models, we should implement policies  using [Pundit](https://github.com/varvet/pundit).

For backend stuff, this can already be implemented. Frontend-wise, I'd wait until my frontend refactor in #68 is done.

pixeldesu:
Same as #809 but for the entire mute rule functionality set. Removed the update action because we don't update `MuteRule`s ever.

Also sets up Pundit (closes #80) to handle authorization for the `destroy` action, instead of an inlined check as in the AJAX action before.

**Testing:**
* Test creating/deleting mute rules


raccube (approved):

raccube (changes requested):
Ignore the previous approve, that was an accident.

The field doesn't empty after submitting.

pixeldesu:
Fixed, the form is now also replaced on submit and I added a Stimulus controller to autofocus the input once that happens. (no, the `autofocus` attribute did not work for me)

raccube (approved):

PR #813, opened 2022-11-20
Add admin dashboard and clean up profile dropdown

pixeldesu:
The profile dropdown includes way too many items, so I created an Administration Panel to house all tools in a sidebar. The `Admin::DashboardController` does nothing at the moment and only displays placeholder text because I don't know what/how to put there.

Additionally, I also moved the moderation panel layout to it's proper position and moved the "Site-wide anonymous blocks" option into the sidebar there as well.

Also, `AnnouncementController` is now housed in `Admin::AnnouncementController` and I fixed the creation of announcements (`data-turbo="false"` was missing)


raccube (changes requested):
When creating an announcement, the flash message shows after navigating away with a missing translastion. (translation missing: `en.admin.announcement.create.success`)

raccube (approved):

PR #826, opened 2022-11-24
Keep "All new notifications" link accessible

Bandie:
Before: 
When the user gets too many notifications, the link to "All new notifications" gets inaccessible.

After:
Keep the link to "All new notifications" accessible by placing it on the top.


pixeldesu (approved):
👍 LGTM. Thanks for your contribution!

PR #841, opened 2022-12-05
Migrate lists to Turbo Frames/Streams

pixeldesu:
Self-explanatory!

**Testing:**
* Test if adding new lists works
* Test if removing lists work
* Test if changing list membership states works


config/routes.rb
```diff
@@ -127,13 +127,15 @@
...
-    post "/destroy_list", to: "list#destroy", as: :destroy_list
-    post "/list_membership", to: "list#membership", as: :list_membership
     post "/subscribe", to: "subscription#subscribe", as: :subscribe_answer
     post "/unsubscribe", to: "subscription#unsubscribe", as: :unsubscribe_answer
   end
 
+  get "/lists", to: "lists#index", as: :lists
+  post "/lists", to: "lists#create", as: :lists_create
+  patch "/lists/:list", to: "lists#update", as: :lists_update
+  delete "/lists/:list", to: "lists#destroy", as: :lists_destroy
```
raccube, line 137: Why not use `resource` here?

config/locales/voc.en.yml
```diff
@@ -27,6 +27,10 @@ en:
     view_in_rails_admin: "View in Rails Admin"
     "y": "Yes"
     "n": "No"
+    members:
+      zero: "0 members"
```
raccube:
```suggestion
      zero: "No members"
```

raccube (approved):

pixeldesu:
There's over 600 commits difference to the `main` branch now, and we also learned some thing regarding Turbo (Streams) so I will redo this PR at a later date, I'll keep the branch however.

PR #852, opened 2022-12-18
Improve notification performance

pixeldesu:
Two things:

* `notification_count` produced an N+1 query (present in a lot of Sentry issues), this should be fixed now
* collecting the notifications for the dropdown doesn't use eager loading for the `target`, this is adjusted now

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -22,7 +22,7 @@ module ApplicationHelper
   def notification_count
     return 0 unless user_signed_in?
 
-    count = Notification.for(current_user).where(new: true).pluck(:id).size
+    count = Notification.for(current_user).where(new: true).count
     return nil unless count.positive?
 
     count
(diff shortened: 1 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -22,11 +22,10 @@ def inbox_count
   def notification_count
     return 0 unless user_signed_in?
 
-    count = Notification.for(current_user).where(new: true)
-    return nil if count.nil?
-    return nil unless count.count.positive?
+    count = Notification.for(current_user).where(new: true).pluck(:id).size
```
raccube: Can we get Rails to generate a query that looks something like `SELECT COUNT(*) FROM notifications WHERE recipient_id = ?` instead of getting a list of IDs?

raccube (approved):

app/helpers/application_helper.rb
nilsding: I guess just something like `....where(...).count` would already do that.  Probably the `.nil?` check caused the extra queries.

nilsding:
```suggestion
    count = Notification.for(current_user).where(new: true).count
```

would generate the nicer `COUNT(*)` query.  The resulting query takes around the same time as the one with pluck(:id).

nilsding (approved):

PR #861, opened 2022-12-21
fix images when not using S3/radosgw

### LINKED ISSUE #822: Make images work in local development environments
With the inclusion of AWS uploading, and presumably the switching to CarrierWave the paths for image uploads broke locally. This should be fixed.

nilsding:
this PR makes local uploads work again.  tried it with my S3 at home, existing and new uploads using that are not affected by this change.

fixes #822


pixeldesu (approved):
LGTM 👍

PR #867, opened 2022-12-23
Various frontend fixes

### LINKED ISSUE #561: Allow scrolling of notifications dropdown on mobile

### LINKED ISSUE #641: Load more button on profile pages is cut off when opened as a PWA on iOS
image

### LINKED ISSUE #863: Question header cuts off space from answers on mobile
Because the question header is fixed, longer questions take away from the space used to display the answers, leaving a small area for doing so.

We should figure out a way of making more space for answers. Maybe allow the question header to be collapsed?

### LINKED ISSUE #865: Comment inputs overlap other content in mobile view
Screenshot from @nilsding:
ima_da6a2b2

pixeldesu:
Guess they had to be fixed someday.

resolves #865 (no overlap)
image

resolves #863 (question is not fixed to top on mobile anymore)
image

resolves #641
_can't actually showcase this because I can't test on a mobile device, tl;dr we now additionally add `env(safe-area-inset-bottom)` to the bottom padding on mobile_

resolves #561 (notification dropdown now has a max height and is scrollable)
image

### DIFF
--- a/app/javascript/styles/components/_comments.scss
+++ b/app/javascript/styles/components/_comments.scss
@@ -22,7 +22,6 @@
   }
 
   &__input {
-    z-index: 99;
     padding-right: 2.5rem;
 
     &.is-invalid {
@@ -32,7 +31,7 @@
 
   &__character-count {
     position: absolute;
-    z-index: 100;
+    z-index: 5;
     right: .5rem;
     top: .5rem;
   }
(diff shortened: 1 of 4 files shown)


nilsding (approved):
tested the fix for #641, works great

raccube (approved):
Looks good. Just two suggestions:
- Could wrap the footer links in a `.container`
- Should we also move the "Show all notifications" link back to the bottom?

nilsding:
I guess the "Show all notifications" link could be at the bottom in the mobile view.  on desktop it makes sense to have it at the top

pixeldesu:
I will address both things in separate PRs, I'm still thinking about the best approach to move the "Show all notifications" link to the bottom without doing duplicate HTML for it, and the container for the footer requires touching a bunch of layouts (where the footer in general is already missing, which should be addressed as well).

PR #868, opened 2022-12-24
Add debugging capabilities to Docker setup

pixeldesu:
_I should have done this way sooner._

Not exactly sure if anyone wants to review this, but instead of launching your Rails app with the debugger, this will just attach to the one running in Docker. Not relevant to (other contributors) that run this locally, but since I work exclusively in containers, I need this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,9 @@ coverage/
 # Curse you, JetBrains!
 .idea/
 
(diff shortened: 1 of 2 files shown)


nilsding (approved):
wird scho passen

raccube (approved):
lol vs code

PR #884, opened 2022-12-27
Track users in anonymous blocks

### LINKED ISSUE #522: Track and block users in anonymous blocks
Add a `user`/`target_user` field to anonymous blocks that tracks the logged in user of an anonymous question.

Combined with #521 this can prevent anonymous questions from specific users even if they change their IP address through VPNs/etc.

pixeldesu:
Finalizing the anonymous blocks to be as effective as they can be, with the options we give. This Pull Request adds an (optional) `target_user` field to anonymous blocks, which is filled as soon as a question has a user.

In the question create use case we then additionally check not just for a matching author identifier, but also a matching user if the question was sent anonymously.

**Testing:** _(Have fun replicating this locally)_
* Ask someone else a question anonymously, while being logged in (User A)
* As the other user, block that anonymous question (User B)
* _User A should change their IP address_
* User A asks User B another anonymous question
* User B does not have a new question in their inbox

If this case is not able to be replicated, there are tests. Otherwise just make sure that asking questions and blocking as before works the same.

Resolves #522

### DIFF
--- a/app/controllers/anonymous_block_controller.rb
+++ b/app/controllers/anonymous_block_controller.rb
@@ -12,7 +12,8 @@ class AnonymousBlockController < ApplicationController
     AnonymousBlock.create!(
       user:       params[:global] ? nil : current_user,
       identifier: question.author_identifier,
-      question:
+      question:,
+      target_user: question.user
     )
 
     inbox_id = question.inboxes.first.id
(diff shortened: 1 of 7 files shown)


nilsding (approved):
tested it locally, looks like it works as designed

raccube (approved):

PR #885, opened 2022-12-27
Silently fail in share worker if user account is suspended

pixeldesu:
_Less errors in Sentry! \o/_

(also we shouldn't handle this on our side since people can also be unsuspended, and having people reconnect their accounts everytime this happens would be a pain...even though it's also questionable)

### DIFF
--- a/app/workers/share_worker.rb
+++ b/app/workers/share_worker.rb
@@ -18,6 +18,10 @@ class ShareWorker
   rescue Twitter::Error::DuplicateStatus
     logger.info "Tried to post answer ##{answer_id} from user ##{user_id} to Twitter but the status was already posted."
     return
+  rescue Twitter::Error::Forbidden
+    # User's Twitter account is suspended
+    logger.info "Tried to post answer ##{answer_id} from user ##{user_id} to Twitter but the account is suspended."
+    return
   rescue Twitter::Error::Unauthorized
     # User's Twitter token has expired or been revoked
     # TODO: Notify user if this happens (https://github.com/Retrospring/retrospring/issues/123)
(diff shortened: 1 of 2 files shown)


app/workers/share_worker.rb
```diff
@@ -14,15 +16,19 @@ def perform(user_id, answer_id, service)
   rescue ActiveRecord::RecordNotFound
     logger.info "Tried to post answer ##{answer_id} for user ##{user_id} to #{service.titleize} but the user/answer/service did not exist (likely deleted), will not retry."
     # The question to be posted was deleted
-    return
+    nil
```
nilsding:
```suggestion
```

app/workers/share_worker.rb
```diff
@@ -14,15 +16,19 @@ def perform(user_id, answer_id, service)
...
   rescue Twitter::Error::DuplicateStatus
     logger.info "Tried to post answer ##{answer_id} from user ##{user_id} to Twitter but the status was already posted."
-    return
+    nil
```
nilsding:
```suggestion
```

app/workers/share_worker.rb
```diff
@@ -14,15 +16,19 @@ def perform(user_id, answer_id, service)
...
+  rescue Twitter::Error::Forbidden
+    # User's Twitter account is suspended
+    logger.info "Tried to post answer ##{answer_id} from user ##{user_id} to Twitter but the account is suspended."
+    nil
```
nilsding:
```suggestion
```

app/workers/share_worker.rb
```diff
@@ -14,15 +16,19 @@ def perform(user_id, answer_id, service)
...
   rescue Twitter::Error::Unauthorized
     # User's Twitter token has expired or been revoked
     # TODO: Notify user if this happens (https://github.com/Retrospring/retrospring/issues/123)
     logger.info "Tried to post answer ##{answer_id} from user ##{user_id} to Twitter but the token has exired or been revoked."
-    return
+    nil
```
nilsding:
```suggestion
```

nilsding (approved):

PR #886, opened 2022-12-27
fix display conditions of profile questions

### LINKED ISSUE #862: Fix display conditions of profile questions
With the addition of moderation unmasking, I tried to implement a feature which only shows public/follower questions in the `/:username/questions` page to users other than the user the profile belongs to (or moderators with enabled unmasking)

This kind of went wrong and the results are now inverted. On my own profile, I can only see direct questions (not all questions) and on other ones only the public ones (which is correct). The first case also happens when moderation unmasking is active.

What actually should be the case:
* If I am on my own profile (regardless of moderation unmasking or not), I can see all questions.
* If I am on a different users profile, I can only see questions that are `direct: false` (e.g. public)
* If I am on a different profile with moderation unmasking active, I can see all questions.

The relevant code for this is in https://github.com/Retrospring/retrospring/blob/main/app/models/user/question_methods.rb

nilsding:
fixes #862 

`cursored_questions` will filter on `direct` if it's non-nil, which is always the case when a bool value is passed to it.

### DIFF
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -53,9 +53,10 @@ class UserController < ApplicationController
 
   def questions
     @title = "Questions"
-    @questions = @user.cursored_questions(author_is_anonymous: false, direct: belongs_to_current_user? || moderation_view?, last_id: params[:last_id])
+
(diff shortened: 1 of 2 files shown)


pixeldesu (approved):
🎉 LGTM

raccube (approved):

Issue #585, opened 2022-07-27
filter out questions from people you're blocking in timelines


nilsding:
scenario:

```cucumber
Given as a fox
When i block user X
And i view my home timeline
Then i should not see any answers to user X's dogshit questions
```

PR #888, opened 2022-12-29
Add user muting

### LINKED ISSUE #16: Muting
Following things should be able to get muted:
* **Keywords:** So questions/answers/inbox entries on feeds don't get shown containing those.
* **Users:** So questions/answers/inbox entries on feeds don't get shown from those.

It also should be able to define a scope on where stuff gets muted, scopes could be:
* Timelines
* Inbox

pixeldesu:
Resolve #16 with the missing piece to muting, users!

User muting currently does two things:
* No (new) questions from them will appear in the inbox
* No notifications for any activity from them will be shown

Questions/Answers are not filtered out from the timelines yet, but this can/should be combined with blocked users in #585.

**Testing:**
* As User A mute a User B
* User B tries to ask User A a question, no inbox entry should be made
* User B follows User A. The follow goes through, but User A is not notified
* User A asks their followers a question, User B responds and User A is not notified of their answer


app/models/user/relationship/mute.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require "errors"
```
nilsding: not `require`d anymore after #889, please rebase

nilsding (approved):

raccube (approved):

PR #890, opened 2023-01-01
Only allow one answer per user

### LINKED ISSUE #391: Two answers from a single user
Description:
Somehow, I ended up answering a question twice, and each of those answers can be returned to inbox. I'm not sure if this can be reliably reproduced.

Steps to reproduce:
Answer a question from the question page, have the page freeze (probably because of network instabilities), go back (or refresh, I don't remember exactly what I did), see that the question wasn't answered, answer it again.

Expected behaviour:
A user shouldn't be able to answer a question twice.

Additional details:
Bromite 100
Android 9

Also, I remember that answering a question at the question page, going back to inbox and trying to answer it again resulted in an error message. But, this time, there were no errors reported.

Screenshots:
1. Question page after deleting the first answer
2. My inbox
3. Question page after answering with something different

Screenshot_20220525-214259_Bromite Screenshot_20220525-214252_Bromite Screenshot_20220525-220403_Bromite

pixeldesu:
quite the simple fix, plus a test!

resolves #391

### DIFF
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -8,6 +8,8 @@ class Answer < ApplicationRecord
   has_many :subscriptions, dependent: :destroy
   has_many :comment_smiles, through: :comments, source: :smiles
 
+  validates :question_id, uniqueness: { scope: :user_id }
+
   # rubocop:disable Rails/SkipsModelValidations
   after_create do
     Inbox.where(user: self.user, question: self.question).destroy_all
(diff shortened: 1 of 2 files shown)


raccube (approved):

app/models/answer.rb
```diff
@@ -8,6 +8,12 @@ class Answer < ApplicationRecord
   has_many :subscriptions, dependent: :destroy
   has_many :comment_smiles, through: :comments, source: :smiles
 
+  # rubocop:disable Rails/UniqueValidationWithoutIndex
+  # This cop is disabled here because there already are questions with
+  # multiple answers. Adding an index would break things database-side
```
nilsding, line 13: We could probably come up with a migration that merges duplicate answers and adds the index.  For now this is good enough

PR #909, opened 2023-01-03
Switch style handling to cssbundling-rails

pixeldesu:
_I had to struggle massively to get this to work._ One part of the journey of removing Webpacker from this project.

Locally: For CSS to be rendered, you now need to run `bin/dev` to use the given Procfile or execute `yarn run build:css` once. In production, everything is hooked into `rails assets:precompile` as usual.

**Testing:**
* Make sure this works locally
* It would be nice if someone had a near-prod environment so we could test a deployment scenario and see if this doesn't completely break the site after a release

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,8 @@
 /tmp
 
 # precompiled assets/packs
+/app/assets/builds/*
+!/app/assets/builds/.keep
 /public/assets
 /public/packs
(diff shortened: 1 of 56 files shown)


raccube (approved):

raccube:
Tested locally in both dev & production mode

PR #922, opened 2023-01-05
remove flag leftovers

nilsding:
Flags are not languages.

We should figure out a better way to display language choices once we need it again.  perhaps a modal with each language as text would be better here


pixeldesu (changes requested):
You can also remove the leftovers in JS.

(`app/javascript/retrospring/features/locales/` and the init  in `app/javascript/packs/application.ts`)

raccube (approved):

pixeldesu (approved):

PR #924, opened 2023-01-05
Prevent WebPush init from attaching events multiple times

raccube:


app/javascript/retrospring/features/webpush/settingsButtons.ts
```diff
@@ -0,0 +1,34 @@
...
+        document.querySelector('[data-action="push-disable"]')?.classList.remove('d-none');
+
+        if (!subscriptionChecked) {
+          checkSubscription(subscription);
+          subscriptionChecked = true;
+          return;
+        }
+      }
+
+      enableBtn?.classList.remove('d-none');
```
pixeldesu:
```suggestion
      } else {
        enableBtn?.classList.remove('d-none');
      }
```

otherwise the enable button always gets shown when you navigate to the page, if you reload on the route it already worked, but not after navigating.

pixeldesu (changes requested):

pixeldesu (approved):

PR #928, opened 2023-01-05
Install and setup jsbundling-rails with esbuild

### LINKED ISSUE #427: Move away from Webpacker to js/cssbundling-rails
_Can't believe we've had this running for just two years until it got retired._

Anyway, back to Sprockets apparently.

Migration guide: https://github.com/rails/jsbundling-rails/blob/main/docs/switch_from_webpacker.md

pixeldesu:
_Fast!_

Business as usual: `yarn install` and `yarn run build` should build the JS, then test the functionality of the site, it should work exactly as before.

Resolves #427


app/assets/javascript/retrospring/i18n.json
```diff
@@ -0,0 +1,134 @@
+{
```
nilsding: que hora es?

nilsding (changes requested):

PR #934, opened 2023-01-06
Make sharing options more obvious

raccube:
Closed:
image
Open:
image


nilsding (approved):

pixeldesu:
I wanted to work on a different system to display those, in general.

Not just for inbox entries, but also the questionbox and question modal. Since the tooling/framework updates were higher prioritized I only kept those in my mind so far.

I don't want to hide options behind a collapsible at all anymore (people will not find them), so here's a quick sketch of what I have in mind:
image

pixeldesu (approved):

PR #935, opened 2023-01-08
Add option to allow long questions

### LINKED ISSUE #311: Remove or greatly increase the length of questions
Currently the character limit for questions seems to be 512 characters which I feel should be greatly increased or even removed. I often find myself wanting to send creators i enjoy longer messages about their content which currently I'm unable to do on retrospring due to this word limit, and I know of others who have the same frustration with the small character limit. I know this issue has been brought up before. But I'm still seeing content creators get complaints about using retrospring due to the limit and asking them to move back to curiouscat because of it.

raccube:
Resolves #311


pixeldesu (changes requested):
This only works partially.

Sending out a question, regardless if it is to followers or to a user directly, causes an error in the frontend (the question is created however).

This comes from the character count reset present in the questionboxes. For the follower modal it can be removed, for the user it needs to be conditional on the element/controller being present.

Also, sending out long questions to followers doesn't work. Long questions are created, but inbox entries are not.

pixeldesu (approved):

PR #936, opened 2023-01-08
Dereference translations

raccube:
Crowdin sadly doesn't support referenced strings.

This also improves a few of the duplicate strings.


pixeldesu (changes requested):
Aside of there being search and replace errors, there is a major problem with this already: Don't duplicate strings. People will have to translate all the duplicates.

Put common used things into the vocabulary.

PR #946, opened 2023-01-10
add spec to verify that generating test locales works

nilsding:
might make sense to have this as part of our test suite since we run this rake task during deployments now


nilsding:
, what

nilsding:
oh!  the code coverage library now sees the rake tasks in lib/ too.  Might as well add another spec for the version task then

pixeldesu (approved):

PR #950, opened 2023-01-12
Add collapsibles for answers and questions

pixeldesu:
Desktop:
image

Mobile:
image

On detail pages the content is always fully shown.

Test with your themes and report any issues. Also check with long new questions if the gradient works for you.

### DIFF
--- a/app/assets/stylesheets/components/_collapse.scss
+++ b/app/assets/stylesheets/components/_collapse.scss
@@ -1,0 +1,94 @@
+@use "sass:map";
+
+.collapsed {
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
+  overflow: hidden;
+
+  &.answerbox__answer-text {
+    -webkit-line-clamp: 12;
+
+    @include media-breakpoint-up('sm') {
+      -webkit-line-clamp: 7;
+    }
+  }
+
+  &.answerbox__question-text {
+    -webkit-line-clamp: 10;
+
+    @include media-breakpoint-up('sm') {
+      -webkit-line-clamp: 5;
+    }
+  }
+}
+
+.collapser {
+  $this: &;
+
+  display: flex;
+  position: sticky;
+  justify-content: center;
+  width: 100%;
+  padding: map.get($spacers, 2);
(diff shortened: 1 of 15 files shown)


app/assets/stylesheets/components/_collapse.scss
```diff
@@ -0,0 +1,94 @@
+@use "sass:map";
+
+.collapsed {
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
+  overflow: hidden;
+
+  &.answerbox__answer-text {
+    -webkit-line-clamp: 12;
```
nilsding, line 9: I really don't like how a `-webkit-` prefix became a "standard" ...

pixeldesu: Nothing I can do about every browser implementing this and not implementing `line-clamp`.

app/helpers/theme_helper.rb
```diff
@@ -36,20 +36,29 @@ def render_theme
     theme.attributes.each do |k, v|
       next unless ATTRIBUTE_MAP.key?(k)
 
-      if k.include?("text") || k.include?("placeholder")
-        hex = get_hex_color_from_theme_value(v)
-        body += "\t--#{ATTRIBUTE_MAP[k]}: #{get_decimal_triplet_from_hex(hex)};\n"
+      if ATTRIBUTE_MAP[k].is_a?(Array)
```
nilsding: Sick Ruby Tricks™ #­54: use [`Kernel#Array`](https://ruby-doc.org/core-3.1.0/Kernel.html#method-i-Array) and you can simply use `.each` regardless whether the thing is an array or not:

```ruby
Array(ATTRIBUTE_MAP[k]).each do |var|
  body += "\t--#{var}: #{get_color_for_key(var, v)};\n"
end
```

nilsding (changes requested):

nilsding (approved):

raccube:
image
I'm happy with it but the button is offset due to the text not filling the width. Would have to make the text wrap around the dropdown caret.

raccube (approved):

PR #951, opened 2023-01-12
Adjust layout of question header to not offset expand button to the side

raccube:


raccube:
This messed up the placement of the user avatar, I had a solution that worked using floats but that's not ideal.

PR #952, opened 2023-01-12
Fix alignment of collapsibles in question headers

pixeldesu:
_(apparently GitHub appended that coauthorship from the last commit in my client, thanks)_

anyway, this fixes the layout weirdness with the alignment of the button now being mostly identical.

image


raccube (approved):

raccube (changes requested):
Just checked again on my phone and got this:

image

raccube (approved):

PR #966, opened 2023-01-17
Refactor capability check into Stimulus controller

pixeldesu:


raccube (approved):
This is fine but i don't really see the point in it being written using Stimulus.

PR #976, opened 2023-01-21
Improve username validation

### LINKED ISSUE #975: Make validation errors for usernames descriptive
"`username` is invalid" is not really helpful, the error should state what exactly is not correct when validating (format, length, etc.)

pixeldesu:
Resolves #975 

Too long username:
image

Wrong format:
image

All the errors:
image

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -63,7 +63,7 @@ class User < ApplicationRecord # rubocop:disable Metrics/ClassLength
   validates :email, fake_email: true, typoed_email: true
   validates :screen_name,
             presence:    true,
-            format:      { with: SCREEN_NAME_REGEX },
+            format:      { with: SCREEN_NAME_REGEX, message: I18n.t("activerecord.validation.user.screen_name.format") },
             length:      { minimum: 1, maximum: 16 },
             uniqueness:  { case_sensitive: false },
             screen_name: true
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -53,15 +53,20 @@ class User < ApplicationRecord # rubocop:disable Metrics/ClassLength
...
 
   before_validation do
     screen_name.strip!
   end
 
   validates :email, fake_email: true, typoed_email: true
-  validates :screen_name, presence: true, format: { with: SCREEN_NAME_REGEX }, uniqueness: { case_sensitive: false }, screen_name: true
+  validates :screen_name,
+            presence:    true,
+            format:      { with: SCREEN_NAME_REGEX },
```
nilsding: would add a message to it too so it's not just "invalid"
```suggestion
            format:      { with: SCREEN_NAME_REGEX, message: "contains invalid characters" },
```

nilsding (changes requested):

app/models/user.rb
pixeldesu: Hmm, fair point, but in combination with the format hint right below it is kinda explaining why it is "invalid", isn't it?

raccube (approved):
Tested locally, happy with the logic. Agree with @nilsding's suggestion to improve the messaging. Although, please put that in a translation.

app/models/user.rb
nilsding: i'd rather have error messages be more explicit, even with the help text below.  e.g. `ü` would count as an alphanumeric character ;-)

nilsding (approved):

PR #979, opened 2023-01-21
Fix various PWA issues

raccube:
- Fixes maskable icon not being used
- Removes `?source=pwa` from PWA start url
- Fixes incorrect description in manifest
- Adds "You are offline" page for PWA

Note: The maskable icons are scaled differently to the regular icons to account for the mask safe area in order to prevent the icon from being cut off.


nilsding (changes requested):
some images are broken?

pixeldesu (approved):
Can't test it locally, but code-wise things look fine.

nilsding (approved):

PR #981, opened 2023-01-21
make Turbo Streams paginate via GET

nilsding:
no more POST!

also got rid of the unused `:page` param while I was at it, seems like a leftover from the dark will_paginate times


raccube (approved):
Tested locally.

pixeldesu (approved):

PR #994, opened 2023-01-23
notifications: only update all new notifications

nilsding:
this should hopefully get rid of some deadlocks we see

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -29,7 +29,7 @@ class NotificationsController < ApplicationController
 
   def mark_notifications_as_read
     # using .dup to not modify @notifications -- useful in tests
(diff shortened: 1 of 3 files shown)


pixeldesu:
in regards of the haml-lint error: we should probably move that into the controller.

nilsding:
well... it at least sent me down a rabbit hole to improve the HAML syntax highlighter of Kate a bit further 😇 

before:

Screenshot 2023-01-23 at 23 26 38

after:

Screenshot 2023-01-23 at 23 22 58

Will take a look at moving it into the controller soon

pixeldesu (approved):

nilsding:
cool, not only did it get rid of the deadlocks, it also made the whole site become way faster too 🚀

PR #998, opened 2023-01-26
Only pass cached requests through service worker

raccube:


raccube:
This was causing issues with ajax requests not being submitted correctly as they would get sent by the service worker rather than the application JS.

PR #1004, opened 2023-01-27
Cache headers for web app manifest based on user theme

raccube:


app/controllers/manifests_controller.rb
```diff
@@ -12,17 +18,19 @@ def show
...
       icons:            webapp_icons,
-      theme_color:      theme_color,
+      theme_color:,
       background_color: mobile_theme_color
     }
   end
 
   private
 
+  def shortcuts = [
```
nilsding, line 30: i love endless method definitions like these :3

nilsding (approved):

PR #1014, opened 2023-01-28
Move fetching of services to controller to avoid n+1 queries

raccube:


app/views/inbox/_entry.html.haml
```diff
@@ -40,9 +40,9 @@
           = render "shared/format_link"
     .card-footer.d-none{ id: "ib-options-#{i.id}" }
       %h4= t(".sharing.heading")
-      - if current_user.services.count.positive?
+      - if @services.count.positive?
```
pixeldesu: Please pass the `@services` into the entry partial so you can use it as `services` and avoid the haml-lint warning.

pixeldesu (changes requested):

app/controllers/inbox_controller.rb
```diff
@@ -8,16 +8,12 @@ class InboxController < ApplicationController
   def show
     find_author
     find_inbox_entries
-
-    if @author_user && @inbox_count.zero?
-      # rubocop disabled because of a false positive
-      flash[:info] = t(".author.info", author: @author) # rubocop:disable Rails/ActionControllerFlashBeforeRender
-      redirect_to inbox_path(last_id: params[:last_id])
```
nilsding, line 15: hmm, I just realised that I wanted to return early in that case ...

app/views/inbox/show.html.haml
```diff
@@ -1,6 +1,6 @@
 #entries
   - @inbox.each do |i|
-    = render "inbox/entry", i:
+    = render "inbox/entry", services: , i:
```
nilsding: dog says:
```suggestion
    = render "inbox/entry", services:, i:
```

nilsding (approved):

pixeldesu (approved):

PR #1017, opened 2023-01-29
Add new raised text and raised accent text theme fields

### LINKED ISSUE #944: separate colours for card tops/bottoms
felt like making a NeXTStep-like theme, but I can't fully pull it off yet.

## current state

the card appendages ("raised content") only allow for a single background colour, and re-use the body text colour.

so my current theme looks like that:

Screenshot 2023-01-09 at 20 25 45

## desired state

first of all, separate text colours for the raised content -- if I want to make the background black it'll be unreadable as the body text colour is black too.

second, have an option to change the tops and bottoms separately if i desire to.  by default that option would be disabled and change the colours for both. 

this would let me change the theme to something like that, making answers look more like NeXTStep windows: 

Screenshot 2023-01-09 at 20 23 36

pixeldesu:
Partially resolves #944.

This adds more options to modify text on raised content, whereas the text on the white/gray sections can now be set independently.

**Before testing:**
* Please create a theme and set `Body text` to a non-default value. A migration is included that sets the `Body text` field value to the new fields, so themes are not completely broken when this feature is released.

**When testing:**
* Don't forget to run `yarn run build` and `yarn run build:css`

**Testing:**
* Set the `Raised text` and `Raised accent text` theme values and check if the preview works and if everything with the styling is okay.


db/migrate/20230129194809_use_body_text_for_raised_texts_in_themes.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class UseBodyTextForRaisedTextsInThemes < ActiveRecord::Migration[6.1]
+  def change
+    execute <<~SQUIRREL
```
nilsding, line 5: Oachkatzl haaßt des!

raccube (approved):
Tested locally.

nilsding (changes requested):
the code works fine and does what it should.

I'm not too sure about the new role of _Body text colour_ though -- it's now literally just the text that's not inside anything that's raised.  It probably would be better to move things around in the theme settings page to make changing the colours of the raised content more prominent.

pixeldesu:
@nilsding I moved the "Raised Content" section to the second position now, between "General" and "Colours"

nilsding (approved):

PR #1029, opened 2023-01-31
Prevent 𝑛+1 on follower/following lists

raccube:


pixeldesu (changes requested):
Not exactly sure if I like this approach.

Userboxes should be able to be used independently, with the option of using something cached like this. (hinting at _"don't use ivars in partials"_ here too)

So please pass the follows down from the `show_follow` template and make it optionally fall back to using `current_user.following?` if it is not passed.

pixeldesu (approved):

Issue #224, opened 2021-12-23
Ability to save and switch between multiple themes


pixeldesu:
_Saw the request on Twitter, and decided to note that down here_

> i kinda wish retrospring let you save themes...
> https://twitter.com/batbite_/status/1474123606078722051

Pretty sure this was part of the initial idea of theming already, but we never got to implement that. Technically, it shouldn't be too hard to do this.

PR #1035, opened 2023-02-01
Refactor user theme CSS to be provided by a stylesheet endpoint

raccube:


app/helpers/theme_helper.rb
```diff
@@ -70,40 +61,26 @@ def theme_color
...
-        else
-          current_user&.theme
-        end
-      else
-        @answer.user.theme
-      end
-    elsif current_user&.theme
-      current_user.theme
+  def active_theme_user
+    user = @user ||= @answer&.user
```
nilsding: would avoid assigning to `@user`, who knows what this can break
```suggestion
    user = @user || @answer&.user
```

app/controllers/user_controller.rb
```diff
@@ -16,6 +20,16 @@ def show
     end
   end
 
+  def show_theme
+    theme = Theme.where(user: User.where("LOWER(screen_name) = ?", params[:username].downcase).select(:id)).first
+    return head :no_content if theme.nil?
```
nilsding: using `unless` here is a bit more idiomatic
```suggestion
    return head :no_content unless theme
```

app/controllers/user_controller.rb
```diff
@@ -1,11 +1,15 @@
...
 class UserController < ApplicationController
-  before_action :set_user
+  include ThemeHelper
+
+  before_action :set_user, except: :show_theme
   before_action :hidden_social_graph_redirect, only: %i[followers followings]
   after_action :mark_notification_as_read, only: %i[show]
 
   def show
+    return show_theme if params[:format] == "css"
```
nilsding, line 12: please add some specs that test this codepath

nilsding (changes requested):

raccube:
Going to use this to implement #224

PR #1037, opened 2023-02-01
Pre-load own mute relationships in follow lists

raccube:
Turns out we were still getting 𝑛+1s here as mutes got missed.

### DIFF
--- a/app/controllers/user_controller.rb
+++ b/app/controllers/user_controller.rb
@@ -19,24 +19,34 @@ class UserController < ApplicationController
   def followers
     paginate_relationships(:cursored_follower_relationships)
     @users = @relationships.map(&:source)
-    own_followings = find_own_relationships(current_user&.active_follow_relationships)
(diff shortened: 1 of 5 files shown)


app/controllers/user_controller.rb
```diff
@@ -19,24 +19,34 @@ def show
   def followers
     paginate_relationships(:cursored_follower_relationships)
     @users = @relationships.map(&:source)
-    own_followings = find_own_relationships(current_user&.active_follow_relationships)
-    own_blocks = find_own_relationships(current_user&.active_block_relationships)
+    own_relationships = find_own_relationships
+    locals = {
+      type:           :friend,
```
pixeldesu:
```suggestion
      type:           :follower,
```

pixeldesu (changes requested):

pixeldesu (approved):

nilsding (approved):

PR #1040, opened 2023-02-01
Prevent 𝑛+1 for notification type counters

raccube:


pixeldesu (approved):

app/views/tabs/_notifications.html.haml
```diff
@@ -9,18 +9,18 @@
   .list-group
     = list_group_item t('.answer'),
       notifications_path('answer'),
-      badge: Notification.for(current_user).where(target_type: 'Answer', new: true).count
+      badge: @counters['Answer']
```
nilsding, line 12: I guess this just returns `nil` if there's no notification with that type yet

nilsding (approved):

PR #1042, opened 2023-02-02
add Containerfile to be used for running Retrospring

nilsding:
This adds a `Containerfile` that will eventually be used to run the web app and workers.  **This is not intended for development**!

It is built from an archive .tar.gz, so we can add another GitHub Actions job that builds it whenever a new tag is pushed.

Using `podman` a new image for a new version (including changes like ruby and bundler versions) can be built automatically like this:

```shell
RETROSPRING_VERSION=2023.0131.1
podman build --build-arg RUBY_VERSION=$(cat .ruby-version) --build-arg BUNDLER_VERSION=$(egrep -A1 "^BUNDLED WITH" Gemfile.lock | tr -d '\n' | awk '{ print $3; }') --build-arg RETROSPRING_VERSION=${RETROSPRING_VERSION} --tag retrospring/retrospring:${RETROSPRING_VERSION} .
```

In the first step we still need to mount the config files into the container, like so:

```shell
podman run --rm -it -v /home/justask/database.yml:/opt/retrospring/app/config/database.yml -v /home/justask/sidekiq.yml:/opt/retrospring/app/config/sidekiq.yml -v /home/justask/justask.yml:/opt/retrospring/app/config/justask.yml -v /home/justask/devise.rb:/opt/retrospring/app/config/initializers/devise.rb -e SECRET_KEY_BASE=seeeecret -e RAILS_LOG_TO_STDOUT=true retrospring/retrospring:2023.0131.1 bundle exec sidekiq
```

Eventually some of the config should be done through ENV variables, probably.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,0 +1,77 @@
+# Container image for a production Retrospring setup
+
+FROM registry.opensuse.org/opensuse/leap:15.4
+
+ARG RETROSPRING_VERSION=2023.0131.1
+ARG RUBY_VERSION=3.1.2
+ARG RUBY_INSTALL_VERSION=0.9.0
+ARG BUNDLER_VERSION=2.3.18
+
+ENV RAILS_ENV=production
+
+# update and install dependencies
+RUN zypper up -y \
+ && zypper in -y \
+      # build dependencies (ruby-install)
+      automake         \
+      gcc              \
+      gdbm-devel       \
+      gzip             \
+      libffi-devel     \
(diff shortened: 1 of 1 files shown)


raccube (approved):
Tested with Podman on openSUSE Tumbleweed

pixeldesu (approved):
Tested with Docker on Windows WSL2 (Ubuntu 20.04)

nilsding:
20.04... wow :)

PR #1044, opened 2023-02-04
Minify assets in production builds

pixeldesu:
Size comparision from dev -> prod builds:

```
application.css: 465kb  -> 376kb
application.js:  1021kb -> 441kb
```

This is kinda hacky because I enhance the jsbundling/cssbundling rake tasks, which means that building effectively runs twice. The build process is incredibly fast though, so it's a bit of a trade here.

### DIFF
--- a/lib/tasks/css.rake
+++ b/lib/tasks/css.rake
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+namespace :css do
+  desc "Build your CSS bundle"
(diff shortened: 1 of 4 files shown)


raccube:
Found a nicer way of doing this: https://github.com/rails/jsbundling-rails/wiki

pixeldesu:
Will re-do differently.

PR #1036, opened 2023-02-01
Move ServicesController -> Settings::ServicesController

pixeldesu:
Resolves #1003

PR #1046, opened 2023-02-05
Implement alternative sharing mechanism

### LINKED ISSUE #1003: Move ServicesController into Settings namespace
All the other controllers are in that namespace already.

**Note:** When adjusting the routes be sure to keep the `/auth` URLs where they are so our configured Twitter application doesn't break.

### LINKED ISSUE #1041: remove support for twitter service connections
post from @TwitterDev: Starting February 9, we will no longer support free access to the Twitter API, both v2 and v1.1. A paid basic tier will be available instead 🧵

source: https://twitter.com/TwitterDev/status/1621026986784337922

----

might be worth to improve manual sharing then -- maybe after answering a question from the inbox, turn the inbox entry into a "Share on XXX" card

pixeldesu:
Instead of using the Twitter API for sharing, we now just go ahead and let users share to services using sharing/intent URLs.

image

Additionally, it allows specifying a custom service sharing URL, so things like Mastodon/Misskey are supported.

image

This PR also:
* removes the `Service` model
* removes the `ServicesController`
* adds a migration to enable `sharing_enabled` on every user that had a `Service`
* removes OmniAuth and other API-related gems (`twitter-text` is kept for text generation)
* removes the `services` section from `justask.yml`

Resolves #1041
Resolves #1003
Superseedes #1036


config/locales/views.en.yml
```diff
@@ -214,10 +214,10 @@ en:
       options: "Sharing Options"
       placeholder: "Write your answer here…"
       sharing:
-        heading: "Sharing"
+        heading: "Share answer on..."
         post_to: "Post to %{service}"
-        none_html: "You have not connected any services yet. Visit your %{settings} to connect one."
-        settings: "service settings"
+        hint_html: "You can customize sharing in your %{settings}"
```
nilsding:
```suggestion
        hint_html: "You can customise sharing in your %{settings}"
```

app/views/settings/sharing/edit.html.haml
```diff
@@ -0,0 +1,20 @@
...
+      = f.form_group :sharing, help: t("activerecord.help.user.sharing_enabled") do
+        = f.check_box :sharing_enabled
+      = f.form_group :sharing, help: t("activerecord.help.user.sharing_autoclose"), class: false do
+        = f.check_box :sharing_autoclose
+
+  .card
+    .card-body
+      %h3= t(".advanced.title")
+      = t(".advanced.body_html")
+      = f.url_field :sharing_custom_url
```
nilsding, line 13: this field should probably have some validation in the model to ensure that `URI(...)` works and is HTTP(S), otherwise it'll break if you enter `asdfkjsaldfj` in there

nilsding (changes requested):
e

app/views/settings/sharing/edit.html.haml
pixeldesu: Validation in the backend has been added. `url_field` will also do HTML-side validation of the form submission (in browsers where that is supported at least).

raccube (changes requested):
Setting a custom sharing URL and removing it does this:
image

config/locales/activerecord.en.yml
```diff
@@ -76,6 +74,9 @@ en:
         privacy_allow_stranger_answers: "Allow other people to answer your questions"
         privacy_noindex: "Prevent search engines from indexing your profile"
         privacy_hide_social_graph: "Hide your social graph from others"
+        sharing_enabled: "Enable sharing"
+        sharing_autoclose: "Automatically close inbox entry after sharing once"
```
raccube: Feels more fitting

```suggestion
        sharing_autoclose: "Automatically hide inbox entry after sharing once"
```

config/locales/views.en.yml
```diff
@@ -214,10 +214,10 @@ en:
       options: "Sharing Options"
       placeholder: "Write your answer here…"
       sharing:
-        heading: "Sharing"
+        heading: "Share answer on..."
         post_to: "Post to %{service}"
-        none_html: "You have not connected any services yet. Visit your %{settings} to connect one."
-        settings: "service settings"
+        hint_html: "You can customise sharing in your %{settings}"
+        settings: "settings"
```
raccube:
```suggestion
        hint_html: "You can customise these options in your %{settings}"
        settings: "sharing settings"
```

raccube (changes requested):

raccube (approved):

nilsding (approved):

PR #1056, opened 2023-02-06
Highlight direct questions in own question lists

### LINKED ISSUE #1038: "Only you can see this" on own direct questions
### Pitch

When viewing the questions tab of your own profile, direct questions should show a "Only you can see this".

### Motivation

A user on Twitter had thought these questions are publicly visible. Having a notice like this would prevent this situation.

raccube:
Resolves #1038

### DIFF
--- a/app/views/moderation/questions/show.html.haml
+++ b/app/views/moderation/questions/show.html.haml
@@ -5,4 +5,4 @@
 
 .container-lg.container--main
   - @questions.each do |q|
-    = render 'shared/question', q: q, type: nil
+    = render "shared/question", q:, type: "moderation"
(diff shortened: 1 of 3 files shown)


pixeldesu (changes requested):
Please localize the strings

nilsding (approved):

pixeldesu (changes requested):
Not sure if I like this UI-wise.

If a lot of questions are direct this adds a lot of additional height to the list (and repeated text). I'd prefer it if (similar to anonymous questions) the icon is present in the header, and the explanation is added as an alert at the top of the question list.

pixeldesu (approved):

PR #1062, opened 2023-02-10
add workflow to automatically build and publish a container image

nilsding:


pixeldesu (approved):

.github/workflows/build-image.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Build and push
+        uses: docker/build-push-action@v4
+        with:
+          build-args: |
+            BUNDLER_VERSION=${{ env.BUNDLER_VERSION }}
+            RETROSPRING_VERSION=${{ env.RETROSPRING_VERSION }}
+            RUBY_VERSION=${{ env.RUBY_VERSION }}
+          context: .
+          file: Containerfile
+          push: ${{ github.event_name != 'pull_request' }}
```
raccube, line 56: Shouldn't this only happen on tag?

nilsding: i feel like having an image for whatever is currently in `main` might be useful at some point in the future.  but i'm not too set on this

raccube (approved):

PR #1064, opened 2023-02-11
Add TurboStreamable concern

### LINKED ISSUE #997: Globally blocking an anonymous user causes an exception if the origin question is in no inbox
...because it tries to delete the inbox entry.

The culprit: https://github.com/Retrospring/retrospring/blob/main/app/controllers/anonymous_block_controller.rb#L20

pixeldesu:
_Toasts in `turbo_stream` responses? For errors even?_

For a long time I kinda struggled with the best ways to implement this and I kinda found a way to achieve this now.

I added a `TurboStreamable` concern which adds toast capability for success/error messages.

Handling of errors is "somewhat" automated. The concern uses the `turbo_stream_actions` method of the parent controller to figure out which actions use Turbo Stream responses, and these actions get wrapped with an error handler that renders a toast for the most common exception types (analogous to `AjaxController`)

It also adds the general `render_toast` method to the invoking controller, adding the ability to send success toast messages. I implemented this into `Settings::MutesController` and `AnonymousBlocksController` (while I was at it, I also fixed #997)

**Testing:**
* Test toasts for anonymous block/mute rule creation and deletion
* Test error toasts
  * _Hint: The easiest way to achieve this one is to create a mute rule and then adjust the action path in the delete button to point to a different ID_

### DIFF
--- a/app/controllers/anonymous_block_controller.rb
+++ b/app/controllers/anonymous_block_controller.rb
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
 class AnonymousBlockController < ApplicationController
+  include TurboStreamable
+
   before_action :authenticate_user!
 
-  def turbo_stream_actions = %i[create destroy]
-
-  include TurboStreamable
+  turbo_stream_actions :create, :destroy
 
   def create
     params.require :question
--- a/app/controllers/anonymous_block_controller.rb
+++ b/app/controllers/anonymous_block_controller.rb
@@ -3,6 +3,10 @@
 class AnonymousBlockController < ApplicationController
   before_action :authenticate_user!
 
+  def turbo_stream_actions = %i[create destroy]
+
+  include TurboStreamable
+
   def create
     params.require :question
 
@@ -21,8 +25,10 @@ class AnonymousBlockController < ApplicationController
 
     respond_to do |format|
       format.turbo_stream do
-        render turbo_stream: turbo_stream.remove("inbox_#{inbox_id}")
+        render turbo_stream: [
           inbox_id ? turbo_stream.remove("inbox_#{inbox_id}") : nil,
+          render_toast(t(".success"))
+        ].compact
       end
 
       format.html { redirect_back(fallback_location: inbox_path) }
@@ -39,7 +45,10 @@ class AnonymousBlockController < ApplicationController
 
     respond_to do |format|
       format.turbo_stream do
-        render turbo_stream: turbo_stream.remove("block_#{params[:id]}")
+        render turbo_stream: [
+          turbo_stream.remove("block_#{params[:id]}"),
+          render_toast(t(".success"))
+        ]
       end
 
       format.html { redirect_back(fallback_location: settings_blocks_path) }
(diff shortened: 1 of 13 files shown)


app/controllers/concerns/turbo_streamable.rb
```diff
@@ -0,0 +1,39 @@
...
+    around_action :handle_error, only: controller.respond_to?(:turbo_stream_actions) ? controller.turbo_stream_actions : []
+  end
+
+  def render_toast(message, success = true)
+    turbo_stream.append("toasts", partial: "shared/toast", locals: { message:, success: })
+  end
+
+  class_methods do
+    def render_toast = render_toast
+  end
```
nilsding: this does not do what you want it to: `controller` is the class itself, which will never respond to the instance method `turbo_stream_actions` (unless you define it as `self.turbo_stream_actions), so the error action will never be applied.

there's a much nicer way to achieve what you want, e.g.:

```suggestion
  class_methods do
    def turbo_stream_actions(*actions)
      around_action :handle_error, only: actions
    end
  end

  def render_toast(message, success = true)
    turbo_stream.append("toasts", partial: "shared/toast", locals: { message:, success: })
  end
```

with that you can include the module right at the beginning of the class, and then just call `turbo_stream_actions :create, :destroy` wherever you need it.

app/controllers/anonymous_block_controller.rb
```diff
@@ -3,6 +3,10 @@
 class AnonymousBlockController < ApplicationController
   before_action :authenticate_user!
 
+  def turbo_stream_actions = %i[create destroy]
+
+  include TurboStreamable
```
nilsding:
```suggestion
  include TurboStreamable

  before_action :authenticate_user!

  turbo_stream_actions :create, :destroy
```

app/controllers/settings/mutes_controller.rb
```diff
@@ -3,6 +3,10 @@
 class Settings::MutesController < ApplicationController
   before_action :authenticate_user!
 
+  def turbo_stream_actions = %i[create destroy]
+
+  include TurboStreamable
```
nilsding:
```suggestion
  include TurboStreamable

  before_action :authenticate_user!

  turbo_stream_actions :create, :destroy
```

nilsding (changes requested):
please add some tests too

pixeldesu:
Okay, so I increased the test coverage.

Rubocop and haml-lint actions seem to fail because something is broken about the Setup Ruby action? Doesn't break just in this PR.

spec/controllers/concerns/turbo_streamable_spec.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe ApplicationController, type: :controller do
```
nilsding:
```suggestion
describe TurboStreamable, type: :controller do
```

nilsding (approved):

raccube (approved):

PR #1071, opened 2023-02-12
Bump rails to 7.0.8

raccube:
Rails Admin has Action Text integration so it assumes Trix is there hence why it has been stubbed


pixeldesu (approved):
Tested this locally and clicked through a bunch of actions and paginations, stuff works fine.

nilsding:
one thing I just noticed: when switching to this branch my session becomes invalid, and therefore will have to sign in again.  wonder if we can keep it compatible somehow

raccube:
Cookie rotation sourced from: https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#key-generator-digest-class-changing-to-use-sha256

raccube:
We should ship MRI 3.2 before we do this

PR #1086, opened 2023-02-16
add some view specs for inbox

nilsding:
this pull request adds our first few view specs to ensure they render fine, even when doing some heavy refactoring.

there's still a lot more to be done of course, but i think `inbox` is a good start.


raccube (approved):
Can we test if paginator (turbo stream) is empty when no more data is available?

pixeldesu (approved):
suprise coverage increase!

PR #1087, opened 2023-02-16
Answerbox caching

raccube:


raccube:
This was a learning exercise for fragment caching but has turned out to be more not the right way of solving this. Although, we could still cache the Markdown render.

raccube:
This was a learning exercise for fragment caching but has turned out to be more not the right way of solving this. Although, we could still cache the Markdown render.

PR #1088, opened 2023-02-16
Refactor subscriptions

raccube:
- Prefetches subscriptions to improve timeline performance
- Refactored subscriptions to use simpler logic/queries

### DIFF
--- a/app/controllers/answer_controller.rb
+++ b/app/controllers/answer_controller.rb
@@ -9,9 +9,13 @@ class AnswerController < ApplicationController
 
   def show
     @answer = Answer.includes(comments: %i[user smiles], question: [:user], smiles: [:user]).find(params[:id])
-    @subscribed = Subscription.where(user: current_user, answer: @answer).pluck(:id)
     @display_all = true
-    mark_notifications_as_read if user_signed_in?
+    @subscribed = []
+
+    return unless user_signed_in?
+
+    @subscribed = Subscription.where(user: current_user, answer: @answer).pluck(:answer_id)
+    mark_notifications_as_read
   end
 
   def pin
--- a/app/controllers/answer_controller.rb
+++ b/app/controllers/answer_controller.rb
@@ -9,6 +9,7 @@ class AnswerController < ApplicationController
 
   def show
     @answer = Answer.includes(comments: %i[user smiles], question: [:user], smiles: [:user]).find(params[:id])
+    @subscribed = Subscription.where(user: current_user, answer: @answer).pluck(:id)
     @display_all = true
 
     if user_signed_in?
--- a/app/controllers/answer_controller.rb
+++ b/app/controllers/answer_controller.rb
@@ -11,17 +11,7 @@ class AnswerController < ApplicationController
     @answer = Answer.includes(comments: %i[user smiles], question: [:user], smiles: [:user]).find(params[:id])
     @subscribed = Subscription.where(user: current_user, answer: @answer).pluck(:id)
     @display_all = true
-
-    if user_signed_in?
-      notif = Notification.where(type: "Notification::QuestionAnswered", target_id: @answer.id, recipient_id: current_user.id, new: true).first
-      notif&.update(new: false)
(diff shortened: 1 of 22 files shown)


app/controllers/timeline_controller.rb
```diff
@@ -32,12 +32,14 @@ def set_lists
 
   def paginate_timeline
     @timeline = yield(last_id: params[:last_id])
-    @timeline_last_id = @timeline.map(&:id).min
+    timeline_ids = @timeline.map(&:id)
+    @timeline_last_id = timeline_ids.min
     @more_data_available = !yield(last_id: @timeline_last_id, size: 1).count.zero?
+    subscribed_answer_ids = Subscription.where(user: current_user, answer_id: timeline_ids).pluck(:answer_id) if user_signed_in?
```
nilsding: `if user_signed_in?` is redundant here, as all of these methods are `:authenticate_user!`

app/controllers/answer_controller.rb
```diff
@@ -9,18 +9,9 @@ class AnswerController < ApplicationController
 
   def show
     @answer = Answer.includes(comments: %i[user smiles], question: [:user], smiles: [:user]).find(params[:id])
+    @subscribed = Subscription.where(user: current_user, answer: @answer).pluck(:id)
```
nilsding: the "subscribe" button never changes unless:

```suggestion
    @subscribed = Subscription.where(user: current_user, answer: @answer).pluck(:answer_id)
```

nilsding: as a guest this still makes a query, i'd rather have it not do one at all in that case.

something like this would be cool:
```ruby
@subscribed = []

...

if user_signed_in?
  @subscribed = Subscription.where ...
  mark_notifications_as_read
end
```

app/controllers/concerns/paginates_answers.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module PaginatesAnswers
+  def paginate_answers
+    answer_ids = @answers.map(&:id)
+    answer_ids += @pinned_answers.pluck(:id) if @pinned_answers.present?
+    @answers_last_id = answer_ids.min
+    @more_data_available = !@user.cursored_answers(last_id: @answers_last_id, size: 1).count.zero?
```
nilsding: I get a `NoMethodError` here on `QuestionController#show`, please fix this and add a test:
```
NoMethodError - undefined method `cursored_answers' for nil:NilClass
    @more_data_available = !@user.cursored_answers(last_id: @answers_last_id, size: 1).count.zero?
                                 ^^^^^^^^^^^^^^^^^:
  app/controllers/concerns/paginates_answers.rb:8:in `paginate_answers'
  app/controllers/question_controller.rb:7:in `show'
  app/controllers/application_controller.rb:23:in `switch_locale'
```

nilsding (changes requested):

raccube:
Subscriptions are a bit of a mess and need to be cleaned up so not going to ship this for now.

nilsding (changes requested):
notifications for comments are broken, the person who commented gets notified instead of whoever subscribed to it?

also, please add specs for everything you fix

app/controllers/concerns/paginates_answers.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module PaginatesAnswers
+  def paginate_answers
+    @answers = yield(last_id: params[:last_id])
+    answer_ids = @answers.map(&:id)
+    answer_ids += @pinned_answers.pluck(:id) if @pinned_answers.present?
+    @answers_last_id = answer_ids.min
+    @more_data_available = !yield(last_id: @answers_last_id, size: 1).count.zero?
+    Subscription.where(user: current_user, answer_id: answer_ids).pluck(:answer_id) if user_signed_in?
```
nilsding: this is a strange api (paginate_answers returns an id of subscriptions?)

instead, have this set an ivar instead of being a return

raccube: e0bce1d

pixeldesu (approved):

PR #1090, opened 2023-02-18
Fix avatar being too large for profile card at portrait tablet size

raccube:


nilsding (approved):

pixeldesu (changes requested):
First of all, don't use queries using `max-width` (the mobile navbar is an exception to this case), it makes it confusing to keep track of responsive changes.

A better approach to fixing this is increasing the column size in the medium viewports, because now the avatar is small, but the column is still super squished together so any amount of text in it looks weird.

Doing something like `col-lg-3 col-md-4` for the profile and `col-lg-9 col-md-8` for the main content should be better.

pixeldesu (approved):

PR #1093, opened 2023-02-18
Update devise to 4.9.0

raccube:


nilsding (changes requested):
tested:

- sign up
- resend confirmation instructions
- sign in without 2FA (signs in, but doesn't redirect to the timeline)
- sign in with 2FA (works fine)
- sign out
- password changing from the user settings 
- password recovery (changes the password, but doesn't redirect to the timeline)

app/controllers/user/sessions_controller.rb
```diff
@@ -42,10 +22,47 @@ def create
...
     respond_with resource, location: after_sign_in_path_for(resource)
   end
-end
\ No newline at end of file
+
+  def prompt_for_2fa
+    session[:user_sign_in_uid] = resource.id
+    sign_out(resource)
+    warden.lock!
+     render "auth/two_factor_authentication"
```
nilsding:
```suggestion
    render "auth/two_factor_authentication"
```

nilsding (approved):

nilsding:
works fine now!

pixeldesu (approved):

PR #1097, opened 2023-02-19
add sharing to telegram

nilsding:
talked with a friend about the new sharing feature, and discovered that telegram has a share endpoint too: https://core.telegram.org/widgets/share

this Pull request implements it 🎉 

Screenshot 2023-02-19 at 20 59 02


raccube (approved):

pixeldesu (approved):
This looks fine to me.

I wonder if we need a setting so people can decide which service they can see in this menu, because if we add some more, this might get quite crowded eventually.

PR #1105, opened 2023-02-20
filter out non-anon questions from blocked/muted users in timelines

### LINKED ISSUE #585: filter out questions from people you're blocking in timelines

nilsding:
~~just an experiment for now~~

attempt to resolve #585 -- please give it some thorough testing!


app/models/user/relationship/mute.rb
```diff
@@ -16,19 +16,41 @@ module Mute
         has_many :muted_by_users, through: :passive_mute_relationships, source: :source
       end
 
+      # Mute an user
```
raccube:
```suggestion
      # Mute a user
```

app/models/user/relationship/block.rb
```diff
@@ -43,6 +57,8 @@ def unfollow_and_remove(target_user)
         inboxes.joins(:question).where(questions: { user_id: target_user.id, author_is_anonymous: false }).destroy_all
         ListMember.joins(:list).where(list: { user_id: target_user.id }, user_id: id).destroy_all
       end
+
+      def cache_key_blocked_user_ids = "#{cache_key}/blocked_user_ids"
```
raccube, line 61: Worth noting that `cache_key` includes the `updated_at` timestamp. Think it's fine here though.

raccube (approved):

app/models/user/relationship/block.rb
nilsding: yes, this is intentional

pixeldesu (approved):
LGTM 👍

PR #1109, opened 2023-02-25
Show notification badges on PWA & cache notification/inbox counters

raccube:


pixeldesu (approved):
Looks good, but fix some of the easier lints please, thanks!

raccube:
Unsure what's going on with that syntax error during lint. Think haml-lint is seeing those two `render`s as one line

PR #1120, opened 2023-03-05
Disable fog in tests

raccube:


pixeldesu (approved):

spec/controllers/settings/profile_picture_controller_spec.rb
```diff
@@ -5,9 +5,16 @@
 describe Settings::ProfilePictureController, type: :controller do
   describe "#update" do
     subject { patch :update, params: { user: avatar_params } }
+
+    before do
+      stub_const("APP_CONFIG", {
```
nilsding: would like to have this toggleable, as i prefer to test these things with a real S3 implementation (like radosgw or s3 itself)

nilsding (approved):

PR #1130, opened 2023-03-07
Mark all notifications as read

raccube:


nilsding (changes requested):
move the buttons back up again please, not only is it less travel to get to the actions when viewing the notifications but it also makes the buttons available on smaller screen heights (just resize your browser a bit and you get what i mean)

the "Mark all as read" button works fine, but now the link to "View all notifications" is broken.  
Screenshot 2023-03-16 at 08 39 10

pixeldesu (approved):
LGTM 👍 (I'll revise the UI for the buttons/links in the notification dropdown in a followup PR)

PR #1149, opened 2023-03-26
Split development and production Docker setup

pixeldesu:
This (possibly) makes it easier to run Retrospring in production using Docker Compose.

The actual `docker-compose.yml` has been removed and replaced with a production-focused `docker-compose.example.yml` which includes comments for important values.

The development setup has been simplified, the `.docker/` directory has been removed, as well as the entrypoint script (as the `command` in the `web` service already removes an existing pid), the customized Dockerfile has been renamed to `Containerfile.dev` and put into the root directory.

To use production or local development docker utilities two new scripts have been added to `bin/`, namely `bin/docker` and `bin/docker-dev` which load the proper compose files for the environment.

_Also added a notice to the `docker-compose.example.yml` file that we do not use this in production ourselves and that we'd appreciate patches if someone has a better idea about this than us._


docker-compose.example.yml
```diff
@@ -0,0 +1,56 @@
...
+
+# NOTE: We don't use Docker in production at Retrospring, so this setup is
+#       provided as-is and might be broken or could be improved.
+#
+#       If you have experience in production docker setups, feel free to
+#       send in some patches!
+
+services:
+  web:
+    image: ghcr.io/retrospring/retrospring:main
```
nilsding, line 11: is there a nice way to specify the version to run from the docker-compose CLI?  this would be preferred for production deployments, as `main` changes all the time

nilsding: is there a nice way to specify the version to run from the docker-compose CLI?  this would be preferred for production deployments, as `main` changes all the time

docker-compose.example.yml
```diff
@@ -0,0 +1,56 @@
+version: "3.7"
+
+# NOTE: We don't use Docker in production at Retrospring, so this setup is
+#       provided as-is and might be broken or could be improved.
+#
+#       If you have experience in production docker setups, feel free to
+#       send in some patches!
+
+services:
```
nilsding, line 10: to be complete this needs to have a `sidekiq` service too

docker-compose.example.yml
```diff
@@ -0,0 +1,56 @@
+version: "3.7"
+
+# NOTE: We don't use Docker in production at Retrospring, so this setup is
+#       provided as-is and might be broken or could be improved.
+#
+#       If you have experience in production docker setups, feel free to
+#       send in some patches!
```
nilsding:
```suggestion
# NOTE: We don't use docker-compose in production ourselves at Retrospring.
#       This setup is provided as-is and might be broken or could be improved.
#
#       If you use have experience with using docker-compose in production (or
#       are adventurous enough to run Retrospring this way yourself), feel free
#       to send in some patches.
```

docker-compose.dev.yml
```diff
@@ -0,0 +1,22 @@
+version: "3.7"
+
+# development docker-compose setup overrides for Retrospring
+# debugging support using rdebug-ide and preconfigured DB credentials
+
+services:
+  web:
```
nilsding, line 7: since it was missing from the "prod" and the previous docker-compose yamls, how do you deal with starting up Sidekiq locally?

bin/docker-dev
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env sh
+
+if [ ! -e "docker-compose.yml" ]; then
+  echo "\033[0;33mdocker-compose.yml not found! Copying example file.\033[0m"
+  cp docker-compose.example.yml docker-compose.yml 
+fi
+
+docker compose -f docker-compose.yml -f docker-compose.dev.yml "$@"
```
nilsding:
```suggestion
exec docker compose -f docker-compose.yml -f docker-compose.dev.yml "$@"
```

also, newline missing

bin/docker
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env sh
+
+if [ ! -e "docker-compose.yml" ]; then
+  echo "\033[0;31mdocker-compose.yml not found! Please create the file from the example file before continuing.\033[0m"
+  exit 1
+fi
+
+docker compose -f docker-compose.yml "$@"
```
nilsding:
```suggestion
exec docker compose -f docker-compose.yml "$@"
```

also, newline missing

nilsding (changes requested):

docker-compose.example.yml
pixeldesu: You have to consider this, the image we push to `main` is the built one from the `Containerfile`, which technically isn't `main` but rather the last tagged version anyway.

Aside of the configuration files, nothing from the repository is mounted in the production setup.

docker-compose.example.yml
pixeldesu: What do we run in that service?

docker-compose.example.yml
nilsding: the `main` tag is built on every push to the `main` branch ;-)

when we release a new build and git-tag them there will be a new container-tag for that version too.

docker-compose.example.yml
nilsding: just `bundle exec sidekiq`

pixeldesu:
Closing this (but keeping the branch for later) because I have the feeling the setup in this state is not going to work out well. I need more experience using Docker-in-Prod to build a proper configuration/building set.

PR #1196, opened 2023-05-07
Move comments to a turbo frame

raccube:


app/views/comment/index.html.haml
```diff
@@ -0,0 +1,3 @@
+= turbo_frame_tag "ab-comments-list-#{a.id}" do
+  %div{ id: "ab-smiles-#{a.id}" }= render "answerbox/smiles", a:
```
pixeldesu: Please don't render the smiles inside the comments frame, they just don't belong here.

Render them in the card footer for now and we'll move them to their own frame later.

pixeldesu (changes requested):
I have not tested this yet, mainly just looked over the code, see one suggestion, additionally:

If you already create a `CommentController` why not make a stream-backed `create` action in it right away?

I feel that would be less trouble than still messing with the `Ajax::CommentController` here.

app/views/comment/index.html.haml
raccube: 13775c6

raccube:
> I have not tested this yet, mainly just looked over the code, see one suggestion, additionally:
> 
> If you already create a `CommentController` why not make a stream-backed `create` action in it right away?
> 
> I feel that would be less trouble than still messing with the `Ajax::CommentController` here.

Will do in a different PR

pixeldesu (approved):
👍 LGTM

Tested on Windows with Vivaldi (Chromium) and Firefox. I'd prefer if there was another Safari on iOS test before this is merged, just to get sure things are working before we release something potentially breaking.

PR #1207, opened 2023-05-09
Prevent links from notifications from being opened in the dropdown frame

raccube:


app/helpers/user_helper.rb
```diff
@@ -23,8 +23,8 @@ def moderation_view?
 
   private
 
-  def profile_link(user)
-    link_to(user.profile.safe_name, user_path(user), class: ("user--banned" if user.banned?).to_s)
+  def profile_link(user, target = nil)
```
nilsding: would be cleaner to have this as a keword argument
```suggestion
  def profile_link(user, target: nil)
```

app/helpers/user_helper.rb
```diff
@@ -12,7 +12,7 @@ def user_screen_name(user, context_user: nil, author_identifier: nil, url: true,
     if url
       return user_path(user) if link_only
 
-      return profile_link(user)
+      return profile_link(user, "_top")
```
nilsding:
```suggestion
      return profile_link(user, target: "_top")
```

nilsding (approved):

pixeldesu (approved):

Issue #511, opened 2022-07-10
Uploading an image with the same filename doesn't crop it.


raccube:
**Steps to reproduce:**

1. Set your profile picture. Do not crop it.
2. Try uploading it again, this time adjust the cropping. After submitting, the cropping will not change.

**Work around**

1. Upload a different picture.
2. Upload the picture you want again
3. Crop adjustment should work as expected.

pixeldesu:
Something we should probably do in general with uploaded files that might fix this: We should either hash the name entirely (so the origin filename is not preserved), or at least prepend a timestamp to the name, so every upload is unique.

nilsding:
I don't think the file name is the problem here, cropping works just fine if you select a crop area (tried this with some kind of S3 storage and local storage).

what I noticed locally is that for some reason the cropper doesn't get initialised properly, it should select the entire picture:

<img width="484" alt="Screenshot 2023-06-16 at 20 14 14" src="https://github.com/Retrospring/retrospring/assets/1809170/710acd02-2424-4cf2-87d1-6aff0b261cc2">

pixeldesu:
This was fixed in [2023.0901.0](https://github.com/Retrospring/retrospring/releases/tag/2023.0901.0)

PR #1240, opened 2023-06-16
Fix race condition in `CropperController`

pixeldesu:
Hopefully solves #511

### DIFF
--- a/app/javascript/retrospring/controllers/cropper_controller.ts
+++ b/app/javascript/retrospring/controllers/cropper_controller.ts
@@ -34,15 +34,19 @@ export default class extends Controller {
 
     if (this.inputTarget.files && this.inputTarget.files[0]) {
       this.readImage(this.inputTarget.files[0], (src) => {
-        this.cropperTarget.src = src;
-
-        new Croppr(this.cropperTarget, {
-          aspectRatio: parseFloat(this.aspectRatioValue),
-          startSize: [100, 100, '%'],
-          onCropStart: this.updateValues.bind(this),
-          onCropMove: this.updateValues.bind(this),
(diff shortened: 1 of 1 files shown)


nilsding (changes requested):
This makes the cropper initialise properly, but it will still use the old crop region as the values are not updated.

app/javascript/retrospring/controllers/cropper_controller.ts
```diff
@@ -34,15 +34,19 @@ export default class extends Controller {
...
-          aspectRatio: parseFloat(this.aspectRatioValue),
-          startSize: [100, 100, '%'],
-          onCropStart: this.updateValues.bind(this),
-          onCropMove: this.updateValues.bind(this),
-          onCropEnd: this.updateValues.bind(this)
+        this.cropperTarget.addEventListener('load', () => {
+          new Croppr(this.cropperTarget, {
+            aspectRatio: parseFloat(this.aspectRatioValue),
+            startSize: [100, 100, '%'],
+            onCropStart: this.updateValues.bind(this),
```
nilsding, line 42:
```suggestion
            onInitialize: (instance) => { this.updateValues(instance.getValue()) },
            onCropStart: this.updateValues.bind(this),
```

app/javascript/retrospring/controllers/cropper_controller.ts
```diff
@@ -34,15 +34,19 @@ export default class extends Controller {
...
+            onCropMove: this.updateValues.bind(this),
+            onCropEnd: this.updateValues.bind(this)
+          });
+        }, {
+          once: true
```
nilsding:
```suggestion
          once: true
```

nilsding (approved):

PR #1301, opened 2023-08-09
Update Lexend fontsource package to version 5

pixeldesu:
So Dependabot doesn't cry around all the time.

> [!IMPORTANT]
> Please check this locally and at best with a production run to see if the fonts get loaded properly.


raccube (changes requested):
Tried this with `RAILS_ENV=production` got a bunch of 404s on the font files

pixeldesu:
_You know what?_ We can just not update to FontSource 5, in the end it's just a package around font assets, nothing critical (unless it would break, but it doesn't)

Issue #100, opened 2020-05-09
Adjusting how Retrospring gets configured


pixeldesu:
Currently, Retrospring gets configured over YAML files, the two prominent ones being `justask.yml` and `database.yml`.

It'd be cool to have an additional/different way to configure the software, especially for development environments like Docker, where having to mess with files before setup is kind of tedious and requires addition script/bootstrap setup.

pixeldesu:
I found this through @nilsding starring it, but probably [configurate](https://github.com/jhass/configurate) is exactly the thing we should use for this.

pixeldesu:
I've been thinking about splitting our Docker setup into a development/production one, but with our current configuration setup it's not that easy, unfortunately.

Configuration-wise, it'd probably be cool if anything that configures service/DB connections would be an environment variable instead, and the Retrospring-specific settings could stay in `justask.yml` still.

We could even use something like https://github.com/bkeepers/dotenv for environment-specific config then.

PR #1328, opened 2023-09-02
use .env and support SMTP configuration

greyidol:
Possible fix for #100. Fixes https://github.com/Retrospring/retrospring/discussions/1009.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@
 .byebug_history
 
 /config/database.yml
+.env
+.env.production
+.env.development
 
 /coverage
 
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 
 /config/database.yml
 .env
-.env.production
 .env.development
 
 /coverage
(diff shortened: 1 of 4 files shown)


pixeldesu (changes requested):
Generally, the addition of configuring email via environment variables is fine.

Something we discussed internally and disagree on is the inclusion of `dotenv` in production. 

Using libraries to manipulate the environment on local machines for ease of development is fine, but in production environments, the environment should be set via deployments/`export`/systemd.

Move the `dotenv-rails` gem into the development/test group section in the Gemfile so it's only installed in those environments.

greyidol:
Should be fixed now. I don't know if I had to do anything to update the lock file though.

greyidol:
I just remembered, the reason I supported the production environment was to ensure that `bundle exec rails c` would work the same way as the unit as started by systemd.

pixeldesu (approved):

pixeldesu:
@greyidol approved and let the lint checks run, could you move `dotenv-rails` into the alphabetically correct position so the lint passes?

raccube (approved):

PR #1370, opened 2023-10-14
Rewrite WebPush enable handler to properly catch errors

### LINKED ISSUE #1369: Inform user why enabling push notifications doesn't work
### Pitch

The website does nothing when clicking the button and it fails. It would be helpful informing what has failed or remind them to check if they're blocking something(?)

### Motivation

EasyList's Notifications sublist is blocking the endpoint for enabling web push.

pixeldesu:
Fixes #1369

**Replication:** (before checking out the PR)
* In (your adblocker of choice) add the "EasyList - Notifications" filter list
* Go into your inbox or the Push Notifications settings and try to enable the notifications
* Nothing happens (no visible error, button states don't change)
* Error in the console about the web request to `/ajax/webpush/key` failing

Above should directly show an error message now when clicking the button to subscribe.

Additionally:
* Fixed the notification style to be an error instead of a success toast.
* Added a null check for the subscription counter (which now threw an error)

### DIFF
--- a/app/javascript/retrospring/features/webpush/enable.ts
+++ b/app/javascript/retrospring/features/webpush/enable.ts
@@ -3,52 +3,53 @@ import I18n from "retrospring/i18n";
 import { showNotification } from "utilities/notifications";
 import { Buffer } from "buffer";
 
-export function enableHandler (event: Event): void {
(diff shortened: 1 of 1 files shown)


raccube (approved):
Tested with "EasyList - Notifications".

PR #1374, opened 2023-10-15
Disable `Metrics/*Length` and `AbcSize` cops

pixeldesu:
These are massive blockers in the development workflow and don't really help much in terms of clean code. Our methods/classes/modules are already relatively short, and complexity-by-number is not as important as actual readability.


nilsding (approved):
might be cool to also remove all `# rubocop:disable Metrics/...` in the code

nilsding (approved):

PR #1388, opened 2023-10-16
Add support for sharing dialog in inbox answer sharing

### LINKED ISSUE #1117: Support share dialog for inbox answer sharing
### Pitch

Some browsers (especially vendored system browsers and on mobile devices) have support to share to other applications on the same device.

### Motivation

Instead of just the predefined services (and the one custom share intent) supporting all user application through this is beneficial not just to the user, but also to us, so we don't need to manually implement support for each.

pixeldesu:
Fixes #1117 

Adds a Stimulus controller for sharing and replaces the "legacy" implementation present in answerboxes.


raccube (changes requested):
**Safari/iOS 17:** When answering a question from your inbox, hitting the Answer button doesn't remove the inbox entry.

raccube (approved):

PR #1397, opened 2023-10-18
Fix user content deletion

nilsding:


pixeldesu:
Works, but it now surfaced that the same is also the case for `Notification`. Procedure of fixing should be exactly the same as with `Inbox` already.

```
[ActiveJob] [ActiveRecord::DestroyAssociationAsyncJob] [ba2aaeeb-5f20-4015-9dda-9ef237a55270] Error performing ActiveRecord::DestroyAssociationAsyncJob (Job ID: ba2aaeeb-5f20-4015-9dda-9ef237a55270) from Sidekiq(default) in 102.54ms: NoMethodError (undefined method `touch' for nil:NilClass

    recipient.touch(:notifications_updated_at)
             ^^^^^^):
/app/app/models/notification.rb:16:in `block in <class:Notification>'
```
https://github.com/Retrospring/retrospring/blob/main/app/models/notification.rb#L16

nilsding:
nice catch, patched this one too

pixeldesu (approved):

nilsding:
ja well…

PR #1400, opened 2023-10-19
Fix lints in database/migration code

pixeldesu:
Self-explanatory.

I got made aware of an interesting lint for migrations however, which we should keep in mind for new migrations: [`Rails/BulkChangeTable`](https://www.rubydoc.info/gems/rubocop/0.61.1/RuboCop/Cop/Rails/BulkChangeTable)


db/schema.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # This file is auto-generated from the current state of the database. Instead
```
nilsding: > This file is auto-generated

how about we exclude this one from the robogarda then?  would be more than annoying to always have to format this file too considering it's auto-generated

pixeldesu: Reverted the `schema.rb` changes since the file is already excluded anyway (oops).

nilsding (approved):

raccube (approved):
Should probably put disable comments on the remaining ones.

PR #1404, opened 2023-10-21
Add `QuestionComponent`

pixeldesu:
_Requires #1403 to be merged first._

De-duplicates question displays across the application.

**Note:** Does not include any tests because...

1. `ApplicationComponent` is just a base component exposing Devise user helpers to the components
2. The above is making it exceptionally hard to unit test components that require user context, because Devise in the end requires Warden/controller context in tests and that's not a given.

I'll try to find a solution for the testing problem later.


app/components/question_component.rb
```diff
@@ -0,0 +1,24 @@
...
+
+class QuestionComponent < ApplicationComponent
+  include ApplicationHelper
+  include BootstrapHelper
+  include UserHelper
+
+  def initialize(question:, context_user: nil, collapse: true, hide_avatar: false, profile_question: false)
+    @question = question
+    @context_user = context_user
+    @collapse = true
```
nilsding: oh yeah rubogarda is right about this one

```suggestion
    @collapse = collapse
```

I wish Ruby had something like Crystal's initialize syntax, where all this wouldn't be necessary:
```crystal
def initialize(@question : Question, @context_user : User?, @collapse = true, @hide_avatar = false, @profile_question = false)
```

app/components/question_component/question_component.en.yml
```diff
@@ -0,0 +1,8 @@
+en:
+  answers:
+    zero: "0 answers"
```
raccube, line 3: Could put something like "No answers" or "Unanswered" here.

raccube (approved):

app/components/question_component/question_component.en.yml
nilsding: Zarro answoors found

pixeldesu: This technically doesn't appear anyway (answer count must be positive to show uppp), it's just added to match all numerical translation values

nilsding: then i would suggest to change it to something more fun

```suggestion
    zero: "zefix, do is jo nix"
```

nilsding (approved):

PR #1411, opened 2023-10-22
Move comment reactions into a Turbo Frame

pixeldesu:
This PR is laying the groundwork for (in Turbo's definition) JS-free modals in Retrospring and replaces the comment reaction modal, that is previously rendered per comment, with a Turbo Frame action.


raccube:
Not really a fan of making a HTTP request / having an endpoint for closing a modal.

pixeldesu:
I get that it's weird, but this is technically reusable code that allows us to close any modal we open with a Turbo Frame. The response is tiny too and doesn't really have any state, so it should be quite fast.

And since it's a website anyway, without offline support, I feel it doesn't make much of a difference.

raccube (approved):

PR #1429, opened 2023-10-29
Refactor `CommentController`

pixeldesu:


app/controllers/comments/reactions_controller.rb
```diff
@@ -1,14 +1,7 @@
...
-  def index
-    answer = Answer.find(params[:id])
-    @comments = Comment.where(answer:).includes([{ user: :profile }, :smiles])
-
-    render "index", locals: { a: answer }
-  end
-
-  def show_reactions
+class Comments::ReactionsController < ApplicationController
+  def show
```
raccube: Think it makes more sense for this to be `index` as it's listing the reactions on a comment rather than showing one

raccube (changes requested):

raccube (approved):

PR #1430, opened 2023-10-29
Improve button styling

pixeldesu:
Buttons are now slightly wider, include an increased border radius, and the font weight is bold. Additionally, the inbox entry layout for buttons is improved as well, as the order/placement was off (since actions are usually on the right).

|         | Before | After |
|---------|--------|-------|
| Mobile  | image | image |
| Desktop | image | image |

### DIFF
--- a/app/assets/stylesheets/_utilities.scss
+++ b/app/assets/stylesheets/_utilities.scss
@@ -30,3 +30,7 @@
 .z-n1 {
   z-index: -1;
 }
+
+.grid-row-1 {
+  grid-row: 1;
+}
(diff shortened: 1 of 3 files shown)


nilsding:
looks neat!

just a small thing in general: could we have a border around the buttons?

pixeldesu:
I'm not sure how to do that yet, but I am thinking of giving users a theme/self option later on to include borders on most UI elements (basically as Bootstrap does it initially), but that requires additional theming variables too, otherwise the borders probably aren't exactly themable yet and it would look incredibly out of place.

In regards to the inbox buttons, on why I decided to do it this way: Using color to denote severity is not accessible and that's why I'm doing a basically "non-button" secondary action, so that delete is still present but not something that draws immediate focus. To make up for the lost "boundaries" without a button background, it includes an icon instead.

It's not perfect yet, but I'll improve the design things over time, this is just an initial step!

nilsding (approved):

raccube (approved):

PR #1441, opened 2023-10-31
Move reactions functionality to Turbo Streams

pixeldesu:


app/controllers/reactions_controller.rb
```diff
@@ -1,9 +1,65 @@
...
     answer = Answer.includes([smiles: { user: :profile }]).find(params[:id])
 
     render "index", locals: { a: answer }
   end
+
+  def create
+    params.require :id
+    params.require :type
+
+    target = params[:type].constantize.find_by!(id: params[:id])
```
nilsding: would extract this `params[:type]` into a private method that's e.g.

```ruby
def create
  params.require :id

  target = target_class.find(params[:id])
  # ...
end

# ...

private

ALLOWED_TYPES = %w[Answer Comment].freeze
private_constant :ALLOWED_TYPES

def target_class
  params.require :type
  raise NameError unless ALLOWED_TYPES.include?(params[:type])

  params[:type].constantize
end
```

raccube (changes requested):
Missing translations
Screenshot 2023-12-09 at 00 24 35

raccube (approved):

PR #1442, opened 2023-11-01
Split up question worker to send to an individual user rather than to all followers

### LINKED ISSUE #810: Split up question worker to send to an individual user rather than to all followers
This will ensure the question is only received once per user in the event of a retry.

raccube:
Closes #810

### DIFF
--- a/app/workers/question_worker.rb
+++ b/app/workers/question_worker.rb
@@ -1,38 +1,40 @@
 # frozen_string_literal: true
 
+# @deprecated This is to be replaced by SendToInboxJob. Remaining here so that remaining QuestionWorker jobs can finish.
 class QuestionWorker
   include Sidekiq::Worker
 
   sidekiq_options queue: :question, retry: false
 
-  # @param follower_id [Integer] user id passed from Devise
+  # @param user_id [Integer] user id passed from Devise
   # @param question_id [Integer] newly created question id
-  def perform(follower_id, question_id)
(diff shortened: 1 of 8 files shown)


pixeldesu (approved):

app/workers/question_worker.rb
```diff
@@ -5,37 +5,34 @@ class QuestionWorker
...
   sidekiq_options queue: :question, retry: false
 
-  # @param user_id [Integer] user id passed from Devise
+  # @param follower_id [Integer] user id passed from Devise
   # @param question_id [Integer] newly created question id
-  def perform(user_id, question_id)
-    user = User.find(user_id)
-    question = Question.find(question_id)
+  def perform(follower_id, question_id)
+    follower = User.includes(:web_push_subscriptions, :mute_rules, :muted_users).find(follower_id)
```
nilsding: one minor concern i have is that during deployment there may be still some `QuestionWorker` jobs that expect the first parameter to be the user id who asked the question, which will eventually end up as them asking the question themselves.

perhaps we can keep the old `QuestionWorker` as a fallback, and rename this worker to something else (`QuestionJob` maybe -- Sidekiq nowadays prefers to use `*Job` instead of `*Worker` https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/worker_compatibility_alias.rb)

nilsding (approved):
looks good overall though

nilsding (approved):

PR #1444, opened 2023-11-05
Fix Missing Translation Issue

### LINKED ISSUE #1423: Title on "View all questions from this user" page is not translated

1. Be an admin/a moderator on Retrospring
2. On a anonymous question, click "Show all questions from this user" in the actions dropdown

### Expected behaviour

The title of the page displays properly.

### Actual behaviour

The title of the page shows the HTML tag for a missing translation.

### Detailed description

_No response_

### Specifications

Retrospring 2023.1019.0

chahmedejaz:
Resolves #1423 
**Issue**
- We are using the wrong translation path for the page title

**Fix**
- This PR updates the correct translation path

**Screenshot**
image

### DIFF
--- a/config/locales/views.en.yml
+++ b/config/locales/views.en.yml
@@ -676,7 +676,7 @@ en:
     questions:
       header:
         title_html: "Viewing all questions from <br><abbr title=%{author_identifier}>%{short}</abbr>"
-      index:
+      show:
         title: "Questions from %{author_identifier}"
   user:
     show_follow:


pixeldesu (changes requested):
Thank you for your (first) contribution!

This generally fixes the problem, but you are solving the symptom rather than the cause in this case.

In case you didn't know, writing `t(".title")` resolves to the `title` locale of the current view, so it unfolds to the full path `moderation.questions.show.title`.

With that in mind, and no other `Moderation::QuestionsController` action actually existing, it is safe to rename the locale in https://github.com/Retrospring/retrospring/blob/main/config/locales/views.en.yml#L679 to match the file name here (just `show`) so we can keep using the shorthand `.title` instead of writing out the full locale path!

chahmedejaz:
> Thank you for your (first) contribution!
> 
> This generally fixes the problem, but you are solving the symptom rather than the cause in this case.
> 
> In case you didn't know, writing `t(".title")` resolves to the `title` locale of the current view, so it unfolds to the full path `moderation.questions.show.title`.
> 
> With that in mind, and no other `Moderation::QuestionsController` action actually existing, it is safe to rename the locale in https://github.com/Retrospring/retrospring/blob/main/config/locales/views.en.yml#L679 to match the file name here (just `show`) so we can keep using the shorthand `.title` instead of writing out the full locale path!

Hey @pixeldesu - Thanks for letting me know. I didn't know that :sweat_smile: 
I've updated as suggested. Please review. Thanks.

pixeldesu (approved):
👍 LGTM

PR #1457, opened 2023-11-06
Fix disabled button issue for the Inbox Question

### LINKED ISSUE #1427: The delete button on inbox items stays disabled after cancelling the deletion operation

1. Go to your inbox
2. On an inbox entry, press "Delete"
3. Press "Cancel"

### Expected behaviour

The "Delete" button is clickable again.

### Actual behaviour

The "Delete" button stays disabled.

### Detailed description

_No response_

### Specifications

Retrospring 2023.1028.1

chahmedejaz:
Resolves: #1427

**Issue**

- In the `deleteEntryHandler`, we are disabling the delete button before the confirmation modal pops up
- On cancel operation, the callback function is not invoked. Hence the button doesn't get re-enabled.
- I believe the package we are using does not invoke the callback function upon cancelation.
- This might have been updated in its recent release if it was the expected behavior before as this code is pretty old.

**The Fix**

- This PR disables the button just before we are making the API call by confirming the deletion.
- Re-enable the button if the API call is unsuccessful

### DIFF
--- a/app/javascript/retrospring/features/inbox/entry/delete.ts
+++ b/app/javascript/retrospring/features/inbox/entry/delete.ts
@@ -7,7 +7,6 @@ import { showNotification, showErrorNotification } from 'utilities/notifications
 
 export function deleteEntryHandler(event: Event): void {
   const element: HTMLButtonElement = event.target as HTMLButtonElement;
-  element.disabled = true;
 
   const data = {
     id: element.getAttribute('data-ib-id')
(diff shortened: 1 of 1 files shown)


app/javascript/retrospring/features/inbox/entry/delete.ts
```diff
@@ -23,11 +22,9 @@ export function deleteEntryHandler(event: Event): void {
     cancelButtonText: I18n.translate('voc.cancel'),
     closeOnConfirm: true
   }, (returnValue) => {
```
chahmedejaz, line 24: - It's observed that we are only invoking this callback if the user confirms the deletion
- It's not invoked when the `cancel` button is pressed. i.e. [line#25](https://github.com/Retrospring/retrospring/pull/1457/files#diff-aaf0f7f045dcad2e26621fb995727e628d3225898748c172e2137e8d5b96aec2R25) is a dead-code and `returnValue` is always `true`
- However, I've still handled the issue in this area as well just in case if I'm missing anything.
- If I've a green signal, I believe it's better to just remove this value and [line#25](https://github.com/Retrospring/retrospring/pull/1457/files#diff-aaf0f7f045dcad2e26621fb995727e628d3225898748c172e2137e8d5b96aec2R25) from the code as well.
- Please suggest the path forward. Thanks.

pixeldesu (approved):

PR #1459, opened 2023-11-07
Pre-emptively fetch answer reaction state

raccube:


pixeldesu (changes requested):
We talked about this in private already, I'll leave the feedback here just so we don't forget about it.

I'm not a fan of this approach because comparing against a pre-fetched set of IDs makes it harder to make independent components out of specific features. It brings a performance improvement, but at the cost of maintainability.

I'm personally not sure how to solve this. If a different way of caching, or using something like

> if a global containing a pre-cached set of IDs is available, prefer that over fetching data yourself

could allow for both cases.

nilsding (approved):

raccube:
@pixeldesu @nilsding I rewrote this using subqueries, so the additional lookup queries aren't required at all.

pixeldesu (approved):

PR #1460, opened 2023-11-09
Fix Error upon Comment

### LINKED ISSUE #1443: Error upon posting comments

1. Post a comment on an answer
2. `TypeError` is thrown.

### Expected behaviour

The subscribe action on the answer should be updated.

### Actual behaviour

`TypeError` is thrown and an error toast is displayed to the user.

### Detailed description

_No response_

### Specifications

Retrospring 20231028.1

Firefox 118.0.2

chahmedejaz:
**Resolves: #1443** 

### Issue
- The `TypeError` is caused by these changes:
  - https://github.com/Retrospring/retrospring/pull/1417/files#diff-dae21ea85421965fd9b72a386fe3461aaca6bc45c3b67d9dd7930c5b01a584dcL3-L10
  - https://github.com/Retrospring/retrospring/pull/1417/files#diff-c52ed63dab8f18456411306029b0efb69cdba2ea74d61c060ef619c02731330cR1-R3
- And in the following lines we are not getting the `sub`, hence the error:
- https://github.com/Retrospring/retrospring/blob/43140105355219ebf23802d9ac506fdb60a562c9/app/javascript/retrospring/features/answerbox/comment/new.ts#L35-L37

### The Fix
- I believe these lines are no longer needed and this PR aims to remove them.
- Please suggest if you think otherwise. Thanks.

### DIFF
--- a/app/javascript/retrospring/features/answerbox/comment/new.ts
+++ b/app/javascript/retrospring/features/answerbox/comment/new.ts
@@ -31,10 +31,6 @@ function createComment(input: HTMLInputElement, id: string, counter: Element, gr
         }
         input.value = '';
         counter.innerHTML = String(512);
-
-        const sub = document.querySelector<HTMLElement>(`[data-action=ab-submarine][data-a-id="${id}"]`);
-        sub.dataset.torpedo = "no"
-        sub.children[0].nextSibling.textContent = ' ' + I18n.translate('voc.unsubscribe');
       }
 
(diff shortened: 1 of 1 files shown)


pixeldesu (approved):
That's good enough. I don't think most users manually subscribe to comments anyway, so until we fix this properly (with refactoring comments to use Turbo as well), this passes.

PR #1461, opened 2023-11-11
[Feature] Copy to Clipboard for Answers

### LINKED ISSUE #1448: Add "Copy to clipboard" sharing option to answers
_Originally referenced from: https://github.com/Retrospring/retrospring/issues/1083#issuecomment-1793472828_

### Pitch

Add the ability to copy the "question - answer [link]" text into the clipboard after answering a question.

### Motivation

This adds the ability to quickly post the sharing text to platforms that don't have support for URL-based sharing intents (like Twitter, Tumblr, etc.)

chahmedejaz:
Resolves: #1448

- This PR adds the feature to copy question in the following format to the clipboard:
  - question - answer [link]
- It also adds the respective specs as well.

Screenshot:
image

image

### DIFF
--- a/app/helpers/social_helper.rb
+++ b/app/helpers/social_helper.rb
@@ -13,4 +13,9 @@ module SocialHelper
       protocol: (APP_CONFIG["https"] ? :https : :http),
     )
   end
+
+  def answer_copy_content(answer)
+    # copy content template: "question - answer [link]"
+    "#{answer.question.content} - #{answer.content} [#{answer_share_url(answer)}]"
+  end
 end
--- a/app/helpers/social_helper.rb
+++ b/app/helpers/social_helper.rb
@@ -13,9 +13,4 @@ module SocialHelper
       protocol: (APP_CONFIG["https"] ? :https : :http),
     )
   end
-
-  def answer_copy_content(answer)
-    # copy content template: "question - answer [link]"
-    "#{answer.question.content} - #{answer.content} [#{answer_share_url(answer)}]"
-  end
 end
--- a/app/javascript/retrospring/controllers/share_controller.ts
+++ b/app/javascript/retrospring/controllers/share_controller.ts
@@ -5,14 +5,12 @@ export default class extends Controller {
   static values = {
     url: String,
     text: String,
-    title: String,
-    copyContent: String
+    title: String
   };
 
   declare readonly urlValue: string;
   declare readonly textValue: string;
   declare readonly titleValue: string;
-  declare readonly copyContentValue: string;
 
   share() {
     let shareConfiguration = {};
@@ -43,9 +41,5 @@ export default class extends Controller {
         this.element.dispatchEvent(new CustomEvent('retrospring:shared'));
       })
       .catch(noop);
-  }
-
-  async copyToClipboard(){
-    await navigator.clipboard.writeText(this.copyContentValue)
   }
 }
--- a/app/javascript/retrospring/controllers/share_controller.ts
+++ b/app/javascript/retrospring/controllers/share_controller.ts
@@ -5,12 +5,14 @@ export default class extends Controller {
   static values = {
     url: String,
     text: String,
-    title: String
+    title: String,
+    copyContent: String
   };
 
   declare readonly urlValue: string;
   declare readonly textValue: string;
   declare readonly titleValue: string;
+  declare readonly copyContentValue: string;
 
   share() {
     let shareConfiguration = {};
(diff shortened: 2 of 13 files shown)


app/helpers/social_helper.rb
```diff
@@ -13,4 +13,9 @@ def answer_share_url(answer)
       protocol: (APP_CONFIG["https"] ? :https : :http),
     )
   end
+
+  def answer_copy_content(answer)
+    # copy content template: "question - answer [link]"
+    "#{answer.question.content} - #{answer.content} [#{answer_share_url(answer)}]"
+  end
```
pixeldesu: You can just use the `SocialHelper::TwitterMethods#prepare_tweet` method, which returns exactly the same format.

The [] around the link in the issue were also just for demonstration, they are not required in the actual output ;)

app/javascript/retrospring/controllers/share_controller.ts
pixeldesu, line 1: I would suggest making a controller specifically for the clipboard feature, so we can reuse it in other places that don't necessarily have to do with sharing.

pixeldesu (approved):
Thanks for the contribution! This is already a good base, I left some suggestions that you should please take a look at!

pixeldesu (changes requested):
I actually meant to request changes instead, sorry! (comments from above still apply)

app/helpers/social_helper.rb
raccube: RDoc comments go above the method instead of inside it.

```suggestion
  # copy content template: "question - answer link"
  def answer_copy_content(answer)
    "#{answer.question.content} - #{answer.content} #{answer_share_url(answer)}"
  end
```

This can also be shortened using end-less method definition like so:
```suggestion
  # copy content template: "question - answer link"
  def answer_copy_content(answer) = "#{answer.question.content} - #{answer.content} #{answer_share_url(answer)}"
```

raccube (changes requested):

app/helpers/social_helper.rb
chahmedejaz: - Thanks for your suggestion. I'd keep this in mind for my future PRs.
- However, as per [this comment](https://github.com/Retrospring/retrospring/pull/1461#discussion_r1390304946), I'll be removing this method. Thanks. :)

chahmedejaz:
Hey @pixeldesu, @raccube - Thanks for your suggestions. I've addressed them. It's ready for review. Thanks :)

chahmedejaz:
> Thanks for your suggestions. I've addressed them. It's ready for review. Thanks :)

Hey @pixeldesu - A soft reminder for this. Please let me know if I'm missing anything here. Thanks.

pixeldesu:
@chahmedejaz I already had a look at the code, but I want to test it locally before the next review. I am quite busy right now so I'll try to review it by the weekend!

chahmedejaz:
> I already had a look at the code, but I want to test it locally before the next review. I am quite busy right now so I'll try to review it by the weekend!

Oh no problem then. :smile:

pixeldesu (changes requested):
So I reviewed this, and the code looks and works fine! Good job 🎉 

After testing, I have two more things that should be addressed:
* The copy action doesn't really have feedback, it feels like I don't really know pressing the button copied the content to my clipboard, we should probably show a notification toast (using `showNotification`) here!*
* The initial feature request was meant to be for the sharing dialog that comes up after answering a question in the inbox. So we should add the button there as well, so people can copy the answer to their clipboard right after answering.

----

*Some notes for adding translations in the JS:
* All strings that should be accessible need to be present in the `frontend.*.yml` file.
* For it to appear in the frontend, you then need to run `rails i18n export` before `yarn run build`, so the new translations are actually packaged.

Sorry again for taking so long with the review, we really appreciate your recent contributions! :)

chahmedejaz:
Thank you so much for the feedback and the guide, @pixeldesu 😄 
No worries, I'll incorporate these as well here.

What's the notification message would you prefer here? I was thinking of:
`Content copied to clipboard`
and on failure (_just to cover all scenarios_):
`Failed to copy content to clipboard`

raccube:
That reads fine in my opinion.

chahmedejaz:
Hey @pixeldesu - I've addressed your [suggestions](https://github.com/Retrospring/retrospring/pull/1461#pullrequestreview-1738529097). Please review and let me know if I'm missing anything. Thanks.

pixeldesu (approved):
👍 Awesome, everything's in order now! Good work!

raccube (approved):

PR #1513, opened 2023-12-11
Set constraints for images in uploader

### LINKED ISSUE #1510: Add validation to file uploads
### Pitch

Add validation to file uploads to limit to a specific file size and specific content types.

### Motivation

Validation makes the process stop early and mention to the user that the file they uploaded isn't supported. It also prevents errors from occuring on the server-side.

### LINKED ISSUE #1511: Strip animation from gif uploads
### Pitch

Remove the animations (frames) from gif uploads and only utilize the first frame of a given upload.

### Motivation

GIFs can't be cropped properly by most image processors (they only tend to catch the first frame anyway).

raccube:
Closes #1510 
Closes #1511

### DIFF
--- a/app/controllers/settings/profile_picture_controller.rb
+++ b/app/controllers/settings/profile_picture_controller.rb
@@ -12,9 +12,10 @@ class Settings::ProfilePictureController < ApplicationController
       text += t(".notice.profile_header") if user_attributes[:profile_header]
       flash[:success] = text
     else
-      flash[:error] = t(".error")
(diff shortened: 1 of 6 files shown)


pixeldesu (changes requested):
Works fine, but it'd be interesting to see if we could show proper validation errors for the improper files instead of just saying "Unable to update profile".

That might lead people to retry over and over, even though the file is just invalid.

pixeldesu (approved):

PR #1580, opened 2024-01-19
Use `attr_accessor` for `Answer` object properties

pixeldesu:
The usage of `attr_writer` for `has_reacted` and `has_subscribed` on `Answer` caused access issues in non-inbox answering of questions, because these properties were not accessible in those cases.


nilsding (approved):

nilsding:
probably the better fix for that would be to remove all occurrences of `has_reacted`/`has_subscribed`, but this will do for now

PR #1594, opened 2024-01-23
Add target user to reports

### LINKED ISSUE #1561: Add target user to reports
### Pitch

This is only technical, nothing user facing.

Add a "target user" field to reports. This makes it easier to group reports to a specific user instead of having to do a complex query across a lot of tables and joins to figure out which report belongs to whom.

### Motivation

This makes it easier for #1562 to be implemented. And probably also is less of a hassle with managing report destruction on account deletion.

pixeldesu:
Fixes #1561

Adds a `target_user(_id)` field to reports, storing the target user, regardless of the report being specifically for a user, answer, question, etc. Includes a migration for currently existing reports to fill the field and of course extending the logic of `Ajax::ReportController#create` to pass the target user into created reports.

The reference field does not include a foreign key and is wholly optional, so reports don't block the account deletion (like it was the case with anonymous blocks).

### DIFF
--- a/db/migrate/20240123182422_add_target_user_to_reports.rb
+++ b/db/migrate/20240123182422_add_target_user_to_reports.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+class AddTargetUserToReports < ActiveRecord::Migration[7.0]
+  def up
+    add_reference :reports, :target_user, null: true, foreign_key: false
+
+    execute <<~SQL.squish
+      UPDATE reports
+      SET target_user_id = users.id
+      FROM users
+      WHERE users.id = reports.target_id AND reports.type = 'Reports::User'
+    SQL
+
+    execute <<~SQL.squish
+      UPDATE reports
+      SET target_user_id = users.id
+      FROM users, comments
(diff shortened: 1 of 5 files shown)


db/migrate/20240123182422_add_target_user_to_reports.rb
```diff
@@ -0,0 +1,39 @@
...
+    execute <<~SQL.squish
+      UPDATE reports
+      SET target_user_id = users.id
+      FROM users, questions
+      WHERE users.id = questions.user_id AND questions.id = reports.target_id AND reports.type = 'Reports::Question'
+    SQL
+  end
+
+  def down
+    remove_reference :reports, :target_user, null: true, foreign_key: { to_table: :users }
```
nilsding: what's that `foreign_key:` for?  line 5 creates it without one

nilsding (approved):

db/migrate/20240123182422_add_target_user_to_reports.rb
pixeldesu: Whoops, leftover from when I had migration issues and thought that a foreign key would solve it, I'll remove that again!

PR #1667, opened 2024-03-10
Added masto verficiation to base.html.haml

thibaultmol:
Sorry If this is incorrect. I'm not used to ruby.
But I think i've added the verification link to your project so that your mastodon profile will show a verification badge next to the website in the bio.
see https://docs.joinmastodon.org/user/profile/#verification


pixeldesu:
Thank you for your contribution!

Retrospring is supposed to be self-hostable, so adding a hardcoded reference to our Fediverse account is kind of counter-productive.

We already support verification with backlinks on profiles (the anchor tags include a `me` rel attribute), so for our own Retrospring account (https://retrospring.net/@retrospring) we could show a verified link on Mastodon.

And given the previous paragraph, a globally configurable instance account in a `link` tag also wouldn't make much sense as it would collide with the profile links if it is parsed first.

Thanks for thinking about us having a verified link on fedi though, that's really thoughtful!

PR #1746, opened 2024-05-25
questions: remove redundant or ungrammatical items

northivanastan:
For a joke on my account, I decided to answer every possible generated question in one post. While doing that, I noticed that a number of questions were redundant - for example, "What is the last thing you ate?" and "What is the last thing you have eaten?" are the same question, except the latter is (I think) grammatically incorrect. Same with "Do you prefer to live in a house or a flat?" and "Would you rather live in a flat or a house?" So I went ahead and removed those questions.

### DIFF
--- a/config/questions/en.yml
+++ b/config/questions/en.yml
@@ -259,12 +259,9 @@
           - song you listened to on repeat
           - thing you:
             - ate
+            - bought
             - did
-            - have:
-              - bought
-              - done
-              - eaten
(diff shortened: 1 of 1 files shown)


pixeldesu (approved):
Thank you for your contribution 🎉 and sorry for taking a while to get back to this. I actually approved pipelines a while ago and then forgot to review and approve.

Things look fine!

PR #1822, opened 2024-08-06
Represent theme colors as hexadecimal in export

pixeldesu:
After toying around with exports I saw that theme colors were sent "raw" (obviously), which most users probably can't do much with. This gives them the _human-readable_ hex colors instead.


spec/lib/use_case/data_export/theme_spec.rb
```diff
@@ -13,45 +14,47 @@
...
+        FactoryBot.create(:theme,
+                          user:,)
       end
     end
 
     it "returns the theme as json" do
       expect(json_file("theme.json")).to eq(
         {
           theme: {
-            id:                 theme.id,
```
nilsding, line 25: no more ids or timestamps?

nilsding (approved):

spec/lib/use_case/data_export/theme_spec.rb
pixeldesu: I mean, we can add them back, just need to do the conditional logic inside the field collecting to skip over them (otherwise it hex's them too).

But to the end user they aren't really that important, I'd say.

PR #1824, opened 2024-08-09
Add ability to disable registrations

pixeldesu:
A little present to the people that decide to host Retrospring themselves.

This is handled via a new configuration option in `justask.yml`:

```yaml
features:
  registration:
    enabled: true
```

Removes all mentions of signing up from the interface (I hope) and redirects away from the sign up/user creation routes if they are tried to be accessed.

----

**Important:** This doesn't come with an invite system, so to create users you either temporarily enable registrations again, or create them manually over RailsAdmin.

### DIFF
--- a/app/controllers/user/registrations_controller.rb
+++ b/app/controllers/user/registrations_controller.rb
@@ -1,9 +1,13 @@
+# frozen_string_literal: true
+
 class User::RegistrationsController < Devise::RegistrationsController
+  before_action :redirect_if_registrations_disabled!, only: %w[create new]
+
   def create
     if captcha_valid?
       super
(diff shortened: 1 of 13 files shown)


pixeldesu:
Rubocop smh:
* Complains about a not defined `new` in before action condition
* Defining the above would literally just be a `new` action with `super`
* Rubocop strips that because that's redundant

----

@nilsding I'm out of clues why the redirect doesn't seem to apply to the create action POST request. Maybe you have an idea.

nilsding:
One thing I noticed: the links in the sign in page still refer to "Sign up" as well

PR #1828, opened 2024-08-11
Add advanced frontpage layout config option

pixeldesu:
The current frontpage is mostly for "marketing" purposes, displaying all the different features Retrospring has. This isn't really required for self-hosted instances. So this Pull Request adds a by default enabled simple layout.

The configuration option added for this is the following:
```yaml
features:
  # Advanced (marketing) frontpage layout
  advanced_frontpage:
    enabled: false
```

| Mobile layout | Desktop layout |
|---------------|----------------|
|image|image|

### DIFF
--- a/app/controllers/about_controller.rb
+++ b/app/controllers/about_controller.rb
@@ -2,9 +2,9 @@
 
 class AboutController < ApplicationController
   def index
-    if Retrospring::Config.advanced_frontpage_enabled?
-      render template: "about/index_advanced"
-    end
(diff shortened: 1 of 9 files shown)


nilsding:
ah, sentry errors are because of some bot activities as indicated by the headers:

```
Accept: application/json, application/xml, text/json, text/x-json, text/javascript, text/xml
User-Agent: RestSharp/1.0.0.0
```

of course there's no json template for that, silly!

PR #2028, opened 2025-02-28
Add customizable source URL

pixeldesu:
If people run forks, adjusting the URL in the template isn't that hard, this just makes it easier to customize this URL.


nilsding (approved):

nilsding (approved):
okay!

PR #2029, opened 2025-02-28
Remove "hardcoded" policies

pixeldesu:
Policies are moved into locales (that was a pain to do) and now include placeholders to use the configured `site_name` instead of Retrospring.

Additionally, the `contact_email` setting used only for WebPush now also finds a use as the contact email present in the Privacy Policy (I was too lazy to add a new config value for this).


nilsding (approved):
passt scho

PR #2030, opened 2025-03-01
Rename `deleted` field to `resolved` in Reports

### LINKED ISSUE #1638: Change `deleted` field in Reports to `resolved`
We don't even have soft deletes (yet), so having a field like this technically doesn't make sense. What is more appropriate in the context of Reports is a `resolved` field.

pixeldesu:
We don't even have soft deletes, so naming the field `deleted` doesn't make much sense.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2024_03_01_203930) do
+ActiveRecord::Schema[7.1].define(version: 2025_03_01_170817) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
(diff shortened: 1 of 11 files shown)


db/schema.rb
```diff
@@ -393,5 +393,6 @@
     t.index ["user_id"], name: "index_web_push_subscriptions_on_user_id"
   end
 
+  add_foreign_key "anonymous_blocks", "users", column: "target_user_id"
```
nilsding, line 396: ?¿?

nilsding (approved):

db/schema.rb
pixeldesu: Rails keeps creating this every time I run migrations. Is it the index on anonymous blocks that creates this foreign key and we just forgot to commit this?

No idea where exactly it'd come from elseway.

nilsding: as long as the tests pass it should be fine ;-) …

pixeldesu: Anonymous blocking and unblocking also works so it should be okay.

PR #2039, opened 2025-03-05
Making website similair to Retrosping

byishanit:
Hello, I wanna use your guys code for making website similar like Retrosping for mini project that i working on, if that's okay to get a access to use your guys codes?


pixeldesu:
Why did you make a PR for this? You could have just opened an issue or a discussion.

But yes, our code is open-source exactly for that reason, so that you can take it and host it yourself.

PR #2040, opened 2025-03-08
Add use case for report creation

### LINKED ISSUE #341: Refactor report functionality into use cases
Read more about this in the [parent issue](https://github.com/Retrospring/retrospring/issues/332)

Functionality location: https://github.com/Retrospring/retrospring/blob/main/app/controllers/ajax/report_controller.rb

Additionally:
I found, or remembered, that the report destroying functionality is residing in a different controller (https://github.com/Retrospring/retrospring/blob/main/app/controllers/ajax/moderation_controller.rb#L47-L65), and we should probably consider moving the `destroy` call there as well.

pixeldesu:
Still on track to add e-mail sending for new reports, didn't want to slap the logic in the user model or the AJAX controller, so I thought the best place would be a UseCase. So...have at it.

This also removes the `User.report` method entirely.

Fixes #341

### DIFF
--- a/app/controllers/ajax/report_controller.rb
+++ b/app/controllers/ajax/report_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class Ajax::ReportController < AjaxController
   def create
     params.require :id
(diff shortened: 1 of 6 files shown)


app/controllers/ajax/report_controller.rb
```diff
@@ -11,35 +13,19 @@ def create
...
-      @response[:message] = t(".unknown")
-      return
+    result = UseCase::Report::Create.call(
+      reporter_id: current_user.id,
+      object_id:   params[:id],
+      object_type: params[:type],
+      reason:      params[:reason],
+    )
+
+    if result[:status] == 201
```
nilsding: I don't think this if/else condition is necessary

`result[:status]` will always be `201`, and the errors raised from the UseCase are caught in the base class https://github.com/Retrospring/retrospring/blob/b4929300fc1923929a679157f6bb27c36b425c23/app/controllers/ajax_controller.rb#L84-L94

lib/use_case/report/create.rb
```diff
@@ -0,0 +1,71 @@
...
+      def object
+        @object ||= case object_type.strip.capitalize
+                    when "User"
+                      ::User.find_by!(screen_name: object_id)
+                    when "Question"
+                      ::Question.find_by(id: object_id)
+                    when "Answer"
+                      ::Answer.find_by(id: object_id)
+                    when "Comment"
+                      ::Comment.find_by(id: object_id)
```
nilsding: using `find_by!` will raise an `ActiveRecord::RecordNotFound` error here, making the `check_object_exists` method obsolete

nilsding (changes requested):

lib/use_case/report/create.rb
nilsding: wait.  on second thought, won't `.find` be enough?

lib/use_case/report/create.rb
```diff
@@ -0,0 +1,73 @@
...
+      def object
+        @object ||= case object_type.strip.capitalize
+                    when "User"
+                      ::User.find_by_screen_name!(object_id)
+                    when "Question"
+                      ::Question.find_by_id(object_id)
+                    when "Answer"
+                      ::Answer.find_by_id(object_id)
+                    when "Comment"
+                      ::Comment.find_by_id(object_id)
```
nilsding: even better:

```suggestion
                    when "User"
                      ::User.find_by_screen_name!(object_id)
                    when "Question"
                      ::Question.find(object_id)
                    when "Answer"
                      ::Answer.find(object_id)
                    when "Comment"
                      ::Comment.find(object_id)
```

then you won't need the `Rails/DynamicFindBy` rubocop exclusion

nilsding (approved):

PR #2044, opened 2025-03-09
Fix build pipeline issues

pixeldesu:
Tests started failing on main after #2042 was merged, while they work just fine locally (and the functionality in itself works fine too)

Let's see if explicitly including `TextHelper` fixes it.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -369,7 +369,7 @@ GEM
     questiongenerator (1.1.0)
     raabro (1.4.0)
     racc (1.8.1)
-    rack (2.2.11)
+    rack (2.2.12)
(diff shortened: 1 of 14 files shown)


pixeldesu:
@nilsding help!

Okay, adjusting the include just shows a different method missing, now we're at

```
undefined method `full_sanitizer' for TimelineController:Class
```

Even worse, here even UseCase::Report stuff fails now? It says `where` is not defined on `Report`. What the hell?

Something is seriously broken and I have no idea what, to be honest. Maybe some wrongly installed version? But clearing Action caches (at least the ones that I _think_ matter) didn't change anything either.

